PR #29, opened 2026-04-18
feat(063): React app runtime with static/vite/node modes and publish CLI

ArianMoeini:
## Summary

Implements spec 063: React app runtime with three modes, per-app session cookies, and a signed publish CLI.

- **Three runtime modes** via `manifest.runtime`: `static` (legacy HTML), `vite` (React SPA from `dist/`), `node` (spawns a child process and reverse-proxies)
- **Process manager** with memory caps (`--max-old-space-size` from `resources.memoryMb`), health checks, idle shutdown, LRU eviction, crash recovery
- **Build orchestrator** with source-hash caching and bounded log buffer
- **Per-app session cookies** HMAC-signed from `MATRIX_AUTH_TOKEN` via HKDF-per-slug, `Path=/apps/{slug}/` scoped
- **Dispatcher** single Hono handler at `/apps/:slug/*` branching on runtime
- **Distribution policy** (`owner-only`/`installable`/`featured`) driven by `listingTrust` + caps, fail-closed
- **`matrix app publish`** CLI: produces signed `source.tar.gz` + `dist.tar.gz` + `dist-hash` + `signature`
- **Install gate** with runtime version negotiation and trusted-path verification
- **Shell integration**: `AppViewer` bootstraps session cookie before iframe navigation, remounts on expiry

## Merge-review blockers (all resolved in the last 4 commits)

1. Dispatcher used to forward `Authorization` + shell-origin cookies into node-runtime children. Since `MATRIX_AUTH_TOKEN` is the HKDF master for `deriveAppSessionKey`, a malicious app could mint session cookies for any slug. Now stripped at the edge with a regression test on `sanitizeCookieHeader`.
2. Shell proxy had an unauthenticated fast-path that rewrote `/api/*`, `/files/*`, `/apps/*` with the admin bearer before Clerk ran. Removed; all requests flow through `withClerk`, and the `platformVerified` branch inside still handles pre-authenticated backend-to-backend rewrites.
3. BuildOrchestrator's stdout/stderr Buffer[] was uncapped — a chatty `pnpm install` or `vite build` could OOM the gateway. Ring-buffered at `MAX_LOG_SIZE`.
4. `serve-static.ts` used `readFileSync`/`statSync`/`existsSync` inline in the Hono handler, blocking the event loop on every asset. Converted to `fs/promises`.

Plus: session bootstrap race fix in `AppViewer`, `extractSlug` regex tightened to reject nested paths (`apps/games/2048/...`), `os-bridge` strips `/files/` prefix from launcher-passed paths, `skipTrailingSlashRedirect` in `next.config.ts` so cookie `Path=/apps/{slug}/` matches, `BUILDER_PROMPT` updated to emit spec-063 `matrix.json`.

## Test plan

- [x] `bun run vitest run tests/gateway/app-runtime tests/gateway/app-runtime*.test.ts tests/shell/os-bridge-open-app.test.ts tests/shell/app-viewer-slug.test.ts` — 229/229 + 54 focused
- [x] Manual: `hello-vite` fixture installs, builds, serves React SPA via `/apps/hello-vite/`
- [x] Manual: `hello-next` fixture spawns node child (PID confirmed under gateway), `/api/health` 200, idle-reap at 35s, respawn after kill, memory cap applied
- [x] Manual: `crash-on-request` fixture: first request 200, process crashes, next request respawns (brief 503 race on immediate retry)
- [x] Manual: 22 legacy apps migrated via `scripts/migrate-apps-063.ts` — all load, Game Center opens nested games (2048 et al.)
- [x] Manual: `matrix app publish` live round-trip produces all 5 bundle artifacts (source.tar.gz, dist.tar.gz, dist-hash, matrix.json, signature)
- [x] `tsc --noEmit` clean on files touched by this branch (pre-existing errors in unrelated `channels/`, `files-tree.ts`, `platform-db.ts`, `social.ts` remain)

## Known follow-ups (non-blocking)

- Manifest's `runtimeState` reports `needs_build` even when the dispatcher successfully serves — stale computation, not a correctness issue.
- Crash-respawn race: ~3s window where immediate retry returns 503 before process-manager cleans up the exited record.
- `server.ts:825,877` return generic `500 "internal"` on invalid manifest — should be typed `422` with detail.


ArianMoeini:
### Code review

Found 2 issues:

1. `BuildOrchestrator.runCommand` spreads the full `process.env` into every `pnpm install` / `vite build` subprocess. `process.env` on the gateway contains `MATRIX_AUTH_TOKEN` — the HKDF master secret used by `deriveAppSessionKey` to sign per-app session cookies — so any community app's build step (postinstall script, Vite plugin) that reads its environment obtains the master key and can mint session cookies for any slug. `process-manager.ts` already uses `safeEnv()` to strip secrets before spawning the runtime child; the build path bypasses it.

https://github.com/HamedMP/matrix-os/blob/9175ec58edb127eb41a22071bbc0c55fc9086357/packages/gateway/src/app-runtime/build-orchestrator.ts#L165-L180

2. `POST /api/apps/:slug/session` reads `await c.req.json()` without a preceding `bodyLimit` guard (root [CLAUDE.md](https://github.com/HamedMP/matrix-os/blob/9175ec58edb127eb41a22071bbc0c55fc9086357/CLAUDE.md) says *\"Use Hono \`bodyLimit\` middleware on every mutating endpoint. Never check Content-Length after the body is already buffered.\"*). Other mutating routes in the same file (wallpaper upload, terminal layout, bridge service) already apply `bodyLimit` — this one is the outlier, so an unbounded POST body is buffered and parsed before any size check runs.

https://github.com/HamedMP/matrix-os/blob/9175ec58edb127eb41a22071bbc0c55fc9086357/packages/gateway/src/server.ts#L866-L897

🤖 Generated with [Claude Code](https://claude.ai/code)

<sub>- If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

ArianMoeini:
### Code review (forced re-review)

No new commits since [the previous review](https://github.com/HamedMP/matrix-os/pull/29#issuecomment-3099900521), but surfacing 6 additional issues on the current tip `9175ec5`:

1. **Publish CLI joins `manifest.slug` and `manifest.version` into paths with no validation.** A malicious/typoed `matrix.json` with slug `"../../.."` or version `"../foo"` escapes the store dir and `writeFile` clobbers files on the publisher's machine. `installApp` validates slug via `SAFE_SLUG`; the publisher side does not. (CLAUDE.md says *"Validate and sanitize all user-supplied values before using in file paths, SQL identifiers, or API URLs. Use `resolveWithinHome` for paths, `SAFE_SLUG` regex for identifiers."*)

https://github.com/HamedMP/matrix-os/blob/9175ec58edb127eb41a22071bbc0c55fc9086357/packages/cli/src/commands/app-publish.ts#L106-L115

2. **`signBundle` uses a literal hardcoded HMAC key (`"matrix-os-publish-key"`).** Key is checked into a public repo, so the signature provides no security — any attacker can forge a \"signed\" bundle. The CLI comment acknowledges this is a stub, but `install-flow.ts`'s verified path accepts it as real.

https://github.com/HamedMP/matrix-os/blob/9175ec58edb127eb41a22071bbc0c55fc9086357/packages/cli/src/commands/app-publish.ts#L67-L75

3. **`installVerifiedApp` silently skips hash verification when `declaredDistHash` is absent.** Docstring promises *\"Fails with BuildError.code = 'hash_mismatch' on divergence\"*, but the hash compare is wrapped in `if (declaredDistHash)`. A community-tier install without a declared hash rebuilds and returns `{ ok: true }` with zero verification — the documented fail-closed guarantee is not upheld.

https://github.com/HamedMP/matrix-os/blob/9175ec58edb127eb41a22071bbc0c55fc9086357/packages/gateway/src/app-runtime/install-flow.ts#L327-L348

4. **Rollback in `installApp` / `installVerifiedApp` only runs on fresh installs.** The pre-emptive `rm(targetDir)` at L174-L177 (and L300-L303) deletes the previous install before the try-block, then the catch rolls back only when `freshInstall === true`. On reinstall failure the user has lost their working install AND is left with a partial/unbootable directory. Use a temp dir + atomic rename instead. Same pattern in `installVerifiedApp` around L300-L354.

https://github.com/HamedMP/matrix-os/blob/9175ec58edb127eb41a22071bbc0c55fc9086357/packages/gateway/src/app-runtime/install-flow.ts#L170-L213

5. **The entire trust-gate / ack-consume flow is unwired.** `install-flow.ts` header comment describes `POST /api/apps/:slug/install` calling `consumeAck` as the terminal step, and `assertInstallAllowed` asserts the *\"single consumer\"* invariant. In reality `server.ts` mounts only `/ack`, `/session`, `/manifest`; `ipc-server.ts:919` still imports `installApp` from the old `app-fork.ts`, not from this new module. `assertInstallAllowed` and the new `installApp`/`installVerifiedApp` are only referenced from tests — ack tokens are minted and peeked but never consumed, and the verified install path has no production caller. (CLAUDE.md says *\"Every IPC tool must resolve its dependency at registration time, not at call time.\"*)

https://github.com/HamedMP/matrix-os/blob/9175ec58edb127eb41a22071bbc0c55fc9086357/packages/gateway/src/app-runtime/install-flow.ts#L1-L13

6. **`--max-old-space-size` injection guard only matches `node ` prefix.** Real app start commands are `next start`, `pnpm start`, `bun run start`, `node_modules/.bin/next start` — none match `startCmd.startsWith(\"node \")`, so the Node memory cap documented in spec 063 is silently skipped for every realistic fixture including `hello-next`'s own `next start`.

https://github.com/HamedMP/matrix-os/blob/9175ec58edb127eb41a22071bbc0c55fc9086357/packages/gateway/src/app-runtime/process-manager.ts#L236-L245

The two findings from the prior review (env leak via `process.env` spread, and `bodyLimit` on `/session`) are resolved — `build-orchestrator.ts:167` now uses `safeBuildEnv`, and `/ack` + `/session` both have `appSessionBodyLimit`.

🤖 Generated with [Claude Code](https://claude.ai/code)

<sub>- If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

PR #30, opened 2026-04-19
feat(066): PR 1 — file sync backend + identity + Cloudflare R2 wiring

HamedMP:
## Summary

- **Identity switch**: gateway keys R2 prefixes off Clerk `claims.sub` (immutable) instead of `MATRIX_HANDLE` (mutable display name). Platform orchestrator injects `MATRIX_USER_ID` into every per-user container so the container-side home-mirror aligns with the API/WS path.
- **Cloudflare R2 wiring**: new `S3_*` env vars forwarded through the Platform orchestrator into every user container; new Sync Storage section in `docs/dev/vps-deployment.md` walks through bucket + API-token provisioning.
- **Single-domain gateway**: `gatewayUrlForHandle` collapses to `https://app.matrix-os.com` for every handle. `/auth/device*` + `/api/auth/device/*` exempt from the container proxy. Legacy `{handle}.matrix-os.com` middleware retired (zero current CLI users in the wild; voice subsystem uses Cloudflare tunnel ingress directly, not the retired platform middleware — flagged in the status file).
- **Client UX**: daemon polls `/api/sync/manifest` until `manifestVersion > 0` (2s interval, 10s per-request + 120s overall timeouts, Content-Type guard, 3-strike hard fail on misconfigured gateway). `matrix login` prints a friendly "sign up first" message on `/api/me` 404 and preserves auth on transient server errors so retry works.

## What's included

| Area | Commits |
|------|---------|
| Docs + runbook | `6ee152b`, `c82cad1` |
| Gateway identity (JWT `claims.sub`) | `084ac2b`, `247d066` (hardening) |
| Platform orchestrator + auth routes | `d34ba1d`, `1894145` (hardening + retirement) |
| Client login + daemon poll-for-manifest | `841f570`, `e45cf5b`, `01148e1`, `06927c3` |
| Home-mirror `MATRIX_USER_ID` alignment | `e5b72dc` |
| Status + group-complete chores | `c82cad1`, `db09982`, `e1d4dd0`, `9c32a71`, `2383c60`, `27a0d14`, `e5b4a17`, `a549c92`, `2c1a469`, `3ab34f0`, `d2a4d0e`, `2e46589`, `4752ec1`, `d3fb70a` |

Full per-group breakdown with acceptance criteria, SHAs, and review findings: [`specs/066-file-sync/pr1-status.md`](specs/066-file-sync/pr1-status.md).
Deploy runbook: [`specs/066-file-sync/deployment-plan.md` § PR 1](specs/066-file-sync/deployment-plan.md).

## Local verification

- `bun run test tests/platform tests/gateway/sync tests/gateway/auth-jwt.test.ts` → **451/451**
- `bun run test packages/sync-client` → **149/149** (2 pre-existing `oauth.test.ts` unhandled-rejection warnings, unrelated to PR 1 — present on `main` before this branch)
- Four specialised review agents + one silent-failure hunter audited the original commits; every critical/high finding was addressed in the fix wave (`247d066`, `e45cf5b`, `01148e1`, `06927c3`, `1894145`).

## Test plan (VPS smoke tests — per deployment-plan.md § PR 1)

Before merge, on the VPS:

- [ ] Cloudflare R2 bucket `matrixos-sync` created + API token provisioned
- [ ] `.env` on VPS populated with `S3_ENDPOINT`, `S3_PUBLIC_ENDPOINT`, `S3_ACCESS_KEY_ID`, `S3_SECRET_ACCESS_KEY`, `S3_BUCKET=matrixos-sync`, `S3_FORCE_PATH_STYLE=false`, `MATRIX_HOME_MIRROR=true`
- [ ] Platform rebuilt + rolling-restarted: `curl -X POST http://localhost:9000/containers/rolling-restart -H "Authorization: Bearer $PLATFORM_SECRET"`
- [ ] `docker exec $CANARY env | grep -E '^(S3_|MATRIX_HOME_MIRROR|MATRIX_USER_ID)='` returns all expected vars
- [ ] Smoke test 1: `curl -sI https://app.matrix-os.com/auth/device?user_code=TEST-1234` → 200, text/html
- [ ] Smoke test 2: `curl -sX POST -H 'content-type: application/json' -d '{"clientId":"smoke-test"}' https://app.matrix-os.com/api/auth/device/code` → 200 with deviceCode/userCode
- [ ] Smoke test 3: container manifest non-empty after boot (home-mirror pushed seed files)
- [ ] Smoke test 4: R2 object prefix is Clerk userId (not handle)
- [ ] Smoke test 5: `matrix login` + `matrix sync ~/matrixos-smoke` pulls container files to laptop
- [ ] Smoke test 6: three-way round-trip (container → laptop, laptop → container) within 5s
- [ ] Smoke test 7: `matrix login` on a Clerk account with no provisioned container → friendly "sign up first" message, no `auth.json` written

## Notes

- **Voice-subsystem caveat**: `packages/platform/src/voice-provisioner.ts` and `packages/gateway/src/voice/tunnel/cloudflare.ts` still build `https://${handle}.matrix-os.com` URLs for Twilio webhooks + voice tunnels. These rely on Cloudflare tunnel ingress mapping the wildcard subdomain to the user container — not the retired platform HTTP middleware — so this PR does not break voice. A full per-handle URL retirement is a spec-046 follow-up.
- **Large-file warning on push**: `apps/menu-bar/.build/index-build/...` 64MB file was already on the branch before this PR (build artifact, unrelated). Worth adding to `.gitignore` in a follow-up.
- **Blocking follow-ups**: none. PR 1 is self-contained.


ArianMoeini:
## Code review + hands-on testing — findings summary

Thanks for the detailed PR and the prior review rounds — it shows. I did a deep pass today (manual + four specialised review agents + running the stack locally end-to-end), focused on the sync backend, device flow, JWT auth, and the sync-client daemon. Below is everything in one place: what's solid, what's blocking, and a concrete path to exercising **laptop → remote → back** before merge.

TL;DR: **7 issues I'd consider merge-blockers**, 5 high-severity, a handful of mediums, plus two integration bugs that only surface when you actually run the stack. All fixes are applied locally on a branch and verified against unit tests (508/508 in scope) + an inside-container S3 smoke test.

---

### Merge blockers

| # | Finding | Location |
|---|---------|----------|
| **B1** | `PLATFORM_JWT_SECRET` / `PLATFORM_JWT_PUBLIC_KEY` / `MATRIX_AUTH_TOKEN` are never piped into user containers. The orchestrator env loop doesn't list them, so `readJwtKeyConfig()` returns `null` inside every container. Combined with `auth.ts:208` (JWT branch is skipped when `jwtKey === null`), every platform-issued device-flow JWT 401s. **JWT auth is broken end-to-end the moment `MATRIX_AUTH_TOKEN` is set in prod.** | `packages/platform/src/main.ts:715-729` |
| **B2** | Presigned PUT URLs don't bind `Content-Length`. Server-side validation caps `file.size ≤ 1GB`, but `getPresignedPutUrl()` builds a `PutObjectCommand` with only `Bucket`+`Key`. A client can declare `size: 1024` and upload 10GB. Bandwidth/storage DoS. Fix: sign `ContentLength` with `unhoistableHeaders: new Set(["content-length"])`, OR use presigned POST with a `content-length-range` policy. | `packages/gateway/src/sync/presign.ts:51`, `r2-client.ts:65` |
| **B3** | Device code is consumed **before** `issueToken()` awaits. The inner transaction deletes the row, then `issueToken` runs. If `getContainerByClerkId()` returns null (common first-login case) or `issueSyncJwt` throws, the device code is gone — CLI sees 500, user has to restart the whole flow. Fix: defer the row delete until after issuance succeeds (idempotent double-redeem is fine; tokens are deterministic per `clerkUserId`). | `packages/platform/src/device-flow.ts:185-208` |
| **B4** | RS256 PEM key loading is broken. `readJwtKeyConfig` does `new TextEncoder().encode(pub)`, but `jose.jwtVerify` with `alg: "RS256"` requires a parsed `KeyLike`/`CryptoKey`, not raw PEM bytes. The HS256 path is exercised in tests; the RS256 path silently explodes the first time anyone rotates to an asymmetric key in prod. Fix: `await importSPKI(pem, "RS256")` with a cache. | `packages/gateway/src/auth-jwt.ts:93-96` |
| **B5** | `checkSharePermission` is dead code — never called by `routes.ts`, `presign.ts`, `commit.ts`, or `home-mirror.ts`. The `/share*` endpoints create DB rows and broadcast WS invites but do NOT enforce cross-user access at the mutation boundary. Every presign is structurally scoped to `caller.sub`, so grantees cannot actually access shared files. Either wire `checkSharePermission` into presign/commit, or explicitly mark sharing as metadata-only in PR 1 (what I did locally). | `packages/gateway/src/sync/sharing.ts:315-343` vs. routes (0 call sites) |
| **B6** | `sync_manifests.user_id` has `REFERENCES users(id) ON DELETE CASCADE`, but **nothing creates the `users` table in the gateway DB**. It only exists in `platform-db.ts` against the `matrixos_platform` database. Gateway runs against `matrixos` — sync init fails with `relation "users" does not exist`. Found this while running `bun run docker` from a clean checkout — sync API is non-functional out of the box. Fix: have the gateway run its own minimal `users` schema, or drop the FK and rely on app-level consistency. | `packages/gateway/src/sync/sharing-db.ts:39,51`, schema missing from gateway bootstrap |
| **B7** | `R2ClientConfig.publicEndpoint` is declared and `S3_PUBLIC_ENDPOINT=http://localhost:9100` is set in compose, but the `r2-client.ts` implementation **never applies it**. Presigned URLs bake the container-internal `http://minio:9000` hostname, which is unreachable from outside the Docker network. **The sync client (CLI daemon, macOS app, browser) cannot upload/download against a Docker-hosted gateway.** Fix: after `getSignedUrl(...)`, swap the URL host to `publicEndpoint` when set. | `packages/gateway/src/sync/r2-client.ts:20` (declared), nowhere (used) |

### High-severity (fix before GA)

- **H6 — `NoSuchKey` substring match is a time bomb.** `manifest.ts:87` accepts `err.name === "NoSuchKey" || err.message.includes("NoSuchKey")`. Any future S3 error whose message contains "NoSuchKey" (e.g. `"Operation NoSuchKey timed out"`) silently returns an empty manifest → next write tombstones every file. Use strict `err.name === "NoSuchKey"` only.
- **H7 — Orphaned R2 blobs on failed PUT→COMMIT.** Only the delete path has post-manifest cleanup. A client dying between `r2.putObject` and `POST /commit` leaves a blob forever. Need a sweeper or R2 lifecycle rule with the `matrixos-sync/` prefix.
- **H8 — `pullFile` reads full R2 body into memory**, no size cap; a slow pull `> SUPPRESS_MS` (5s) can overwrite a concurrent user edit in the container. Stream + cap via `maxPushBytes`.
- **H9 — JWT validation failures log at `console.debug`.** Effectively silent in production log levels. A rotated `PLATFORM_JWT_SECRET` would 401 every device with zero operator signal. Use `console.warn`.
- **H10 — `initialPush` holds the advisory lock** while uploading every file serially. On a fresh container with 10k seed files this blocks every commit for minutes. Batch + incremental manifest writes.

### Medium

- M1: Unbounded `enqueue` chain in `home-mirror` — many concurrent peer broadcasts can stack arbitrary depth (`home-mirror.ts:458-481`).
- M3: `home-mirror.normalizeRelativePath` passes literal `"home-mirror"` as `userId` to `resolveWithinPrefix` — bypasses any future per-user path rules (`home-mirror.ts:116`).
- M4: Commit cap-detection uses `/cap|limit|50/i` message regex — brittle; use a typed `ManifestCapExceededError`.
- M5: `ManifestEntry` allows illegal states (`deleted: true` without `deletedAt`; `CommitFile.action === "delete"` still requires a content hash that gets written into the tombstone). Discriminated union would make this unrepresentable.
- M6: `PresignFile` with optional `size` means an omitted `size` bypasses the 1GB cap entirely at the schema layer. Split into `PresignPut` (hash+size required) / `PresignGet`.
- M7: Auth-routes rate-limiter bumps LRU recency on **throttled** requests too — an attacker hammering the limiter keeps their bucket alive forever and evicts legit entries (`platform/src/auth-routes.ts:42-54`).
- M9: `SyncJwtClaims.handle` required but Clerk usernames are mutable — rename invalidates all tokens silently. Either optional+re-derive on verify, or revoke tokens on rename.
- M10: `SyncJwtClaims` + signing code is duplicated across `gateway/src/auth-jwt.ts` and `platform/src/sync-jwt.ts` and has already drifted (platform doesn't check `sub.length > 0`). Move to a shared package.

### Things done right 👏

Worth calling out — these show the prior review rounds landed:

- `timingSafeCompare` padding in `auth.ts:90-99` correctly eliminates the length-oracle bug (with the right comment explaining why).
- `expectedHandle` mismatch check in JWT validation prevents cross-tenant token reuse.
- CSRF double-submit with `SameSite=Strict; HttpOnly; Path=/auth/device` + `timingSafeEqual`-compared bodies.
- `withAdvisoryLock` serialises manifest mutations per-user — the right primitive.
- Startup assertion at `server.ts:361` refuses to boot with `MATRIX_HOME_MIRROR=true` but no `MATRIX_USER_ID`.
- `AbortSignal.timeout(10_000)` on every S3 op.
- Explicit caps + LRU on every in-memory Map (`MANIFEST_FILE_CAP`, `MAX_PEERS_PER_USER`, `MAX_USERS`, `recentlyWritten`).
- `waitForManifest` in daemon checks Content-Type — catches captive-portal / misconfigured proxy.

---

### Fixes applied locally

I've implemented fixes for **B1, B2, B3, B4, B5, H6, H9, M2, M3, M4, M7, and three LOW items** on a review branch. 18 files, +315/-74. Unit tests:

```
tests/gateway/sync/  tests/platform/  tests/gateway/auth-jwt.test.ts  →  508/508 passing
packages/sync-client                                                  →  167/167 passing
```

**B2 (Content-Length signing) verified empirically** against MinIO from inside the dev container:

```
Legit 15-byte upload (size matches signature):           HTTP 200 ✅
Same signed URL, 5MB body (size-bypass attempt):         HTTP 403 SignatureDoesNotMatch ✅
Presigned URL contains: X-Amz-SignedHeaders=content-length;host ✅
Schema rejects action:"put" without size:                400 Validation error: expected number ✅
```

I can open this as a follow-up PR against `main` once #30 merges, or if it's easier I can push the commits to `066-file-sync` directly. Let me know which you prefer.

---

## How to actually test this end-to-end (laptop → VPS)

The 7 VPS smoke tests in the PR body are good as a post-deploy sanity check. But the PR *also* needs a pre-merge "can real files actually round-trip from my laptop to a remote machine and back" test. Here's a concrete three-tier plan.

### Tier 1 — All-local smoke (what I ran today, ~10 min)

**Goal:** confirm sync HTTP API + home-mirror work on one host.

```bash
cd matrix-os
pnpm install
bun run docker              # brings up postgres + minio + gateway + minio-init

# Wait for "[sync] Sync API initialized (S3 endpoint: http://minio:9000 )"
# If you see "[sync] Failed to initialize sync: relation \"users\" does not exist",
# you hit B6 — workaround:
docker exec matrix-os-postgres-1 psql -U matrixos -d matrixos -c \
  "CREATE TABLE IF NOT EXISTS users (id TEXT PRIMARY KEY, handle TEXT UNIQUE); \
   INSERT INTO users (id, handle) VALUES ('dev', 'dev') ON CONFLICT DO NOTHING;"

# Run the presign/commit cycle from sync-testing.md. Note: because of B7
# the presigned URLs will point at minio:9000 (unreachable from host), so
# the full curl demo only works from INSIDE the container:
docker compose -f docker-compose.dev.yml exec dev bash -lc '
  echo -n "hello" > /tmp/t.txt
  HASH="sha256:$(sha256sum /tmp/t.txt | cut -d" " -f1)"
  SIZE=$(wc -c </tmp/t.txt | tr -d " ")
  curl -s -X POST -H "Content-Type: application/json" \
    -d "{\"files\":[{\"path\":\"t.txt\",\"action\":\"put\",\"hash\":\"$HASH\",\"size\":$SIZE}]}" \
    http://localhost:4000/api/sync/presign
'

# Browse MinIO to see objects at http://localhost:9101 (matrixos / matrixos123)
```

**What this proves:** sync API works, manifest commits work, home-mirror pushes the seed home to S3, identity keying on `MATRIX_USER_ID` is correct. **Does NOT prove** real laptop→remote sync.

### Tier 2 — LAN two-host test (closest to real VPS without paying for one, ~30 min)

**Goal:** actually verify the CLI daemon on a laptop can sync to a gateway on a different machine.

**Gotcha — B7 must be fixed first.** Without the `publicEndpoint` rewrite, the daemon will receive presigned URLs containing the gateway host's internal `minio:9000` and fail to connect. The fix is a one-liner in `r2-client.ts`:

```ts
// After getSignedUrl() in each method:
if (publicEndpoint) {
  return signed.replace(/^https?:\/\/[^/]+/, publicEndpoint);
}
return signed;
```

Once that's in:

- **Machine A (the "VPS"):** any spare Mac, Linux box, or $5/mo droplet on your LAN.

  ```bash
  # On Machine A, open ports 4000 + 9100 to your LAN:
  git clone <matrix-os> && cd matrix-os
  pnpm install && bun run docker

  # (Or use a real VPS — works the same.)
  # Note Machine A's LAN IP:
  ip addr | grep inet | grep -v 127.0.0.1 | head -1
  # → e.g., 192.168.1.50

  # Override S3_PUBLIC_ENDPOINT in compose so presigned URLs use the LAN IP:
  # In docker-compose.dev.yml:  S3_PUBLIC_ENDPOINT=http://192.168.1.50:9100
  # Apply B6 + users-seed workaround from Tier 1.
  ```

- **Machine B (the laptop):**

  ```bash
  cd matrix-os/packages/sync-client
  bun install && bun run build

  node --import tsx ../../bin/matrixos.ts login --dev

  # Edit ~/.matrixos/config.json to point at Machine A:
  #   "gatewayUrl": "http://192.168.1.50:4000",
  #   "platformUrl": "http://192.168.1.50:9000"

  mkdir -p ~/matrixos-test
  node --import tsx ../../bin/matrixos.ts sync ~/matrixos-test
  # Leave this running in a terminal.
  ```

- **Verify the round-trip:**

  ```bash
  # On Machine B:
  echo "laptop → VPS" > ~/matrixos-test/laptop.txt
  sleep 5

  # On Machine A:
  docker exec matrix-os-dev-1 ls -la /home/matrixos/home/
  # Expect to see laptop.txt

  # Reverse:
  docker exec matrix-os-dev-1 sh -c 'echo "VPS → laptop" > /home/matrixos/home/vps.txt'
  sleep 5

  # On Machine B:
  ls ~/matrixos-test
  # Expect: laptop.txt AND vps.txt
  ```

**What this proves:** the whole point of the PR. A laptop with the CLI daemon genuinely round-trips files to a remote gateway across a real network. Catches B7 (trivially) and any gateway↔R2 networking issues the single-host test hides.

### Tier 3 — Full production (real VPS + Cloudflare R2 + Clerk)

The PR body's 7 smoke tests cover this well. Key additional things to check after deploy:

1. `docker exec <user-container> env | grep -E '^(PLATFORM_JWT|MATRIX_AUTH|MATRIX_USER_ID|S3_)='` → **all present** (tests B1 in real prod).
2. Run the Tier 2 CLI round-trip against the VPS (`matrix login` with real Clerk flow, not `--dev`). Catches any identity-chain bugs B1 left behind.
3. **Size-bypass attack via presign:** have a test user declare `size: 1024` via `/api/sync/presign`, then attempt to upload 100MB with the returned URL. Expect `403 SignatureDoesNotMatch` (tests B2 fix in prod).
4. **Double-device-flow:** complete `matrix login` on a Clerk account that has no container provisioned. Without B3, you'll see `expired_token` and be forced to restart. With B3, you'll see the friendly "sign up first" message from `login.ts`.
5. Sentry / alerting on `[auth] JWT validation failed` warnings — lets you catch a rotated-secret incident before users report it (tests H9).

### What I'd gate the merge on

- **B1, B2, B5** are security-critical — block merge.
- **B3** is a UX-breaking bug that will page you on day 1 — block merge.
- **B6, B7** block anyone running the PR locally from a clean checkout — essentially block merge because nobody can validate the PR end-to-end without manual DB surgery.
- **B4** is latent — block if you plan to rotate to RS256 in the next quarter; fine as a follow-up otherwise.
- Everything else → follow-up PRs are fine.

Happy to push the applied fixes — say the word and I'll open a follow-up PR (or force-push, if you'd rather keep the history on this branch).

HamedMP:
@greptileai review this fully again

PR #89, opened 2026-05-10
fix(tests): restore repo-wide unit baseline

HamedMP:
## Summary
- Restore clean Vitest source resolution for the kernel workspace package without requiring prebuilt dist artifacts, including the E2E Vitest config.
- Update stale and timing-sensitive gateway, platform, and shell unit fixtures to match current behavior under full-suite load.
- Fix pattern-scan violations in system update and customer VPS fetch probes while preserving timeout behavior.

## Tests
- bun run typecheck
- bun run check:patterns
- bun run test:e2e
- bun run test

## Invariants
- Source of truth: production code paths remain unchanged for persistence; tests now align with current docs, matrix-update launcher behavior, current shell defaults, and current app manifest schema. Kernel runtime exports still point at dist, with TypeScript-only source type conditions and Vitest configs aliasing clean-checkout test imports to source.
- Lock/transaction scope: no database transaction scope changed. Home-mirror tests continue to assert manifest advisory-lock use and upload cleanup on manifest write failure.
- Acceptable orphan states: existing local-upload cleanup behavior is preserved and now tested with load-tolerant waits; customer VPS probe/deploy failures continue to return coarse failed/false results.
- Auth source of truth: sync auth remains JWT claims first and legacy bearer fallback second. The tampered JWT test now mutates a significant signature byte so forged claims cannot pass accidentally.
- Deferred scope: no MAT-6 browser implementation changes, no production rollout changes, and no broader refactors beyond keeping the repo-wide unit and pre-PR gates green.


HamedMP:
Closing because the Linear issue for branch codex/mat-9-green-tests entered a terminal state without merge.

PR #92, opened 2026-05-11
fix(gateway): align update manifest URL with publish script path

ArianMoeini:
Gateway checked /system-bundles/channels/{channel}.json but publish script writes to /system-bundles/{channel}/manifest.json on R2. Users never see 'Update available' in Settings for CI-published bundles. Aligns the gateway path to match.


HamedMP:
Closing this stale/conflicting PR as part of cleanup. If the manifest URL fix is still needed, it should be recreated from current main.

PR #93, opened 2026-05-11
feat(observability): add PostHog error tracking

HamedMP:
## Summary
- Add a shared @matrix-os/observability package for PostHog server/client configuration, Hono exception capture, Next.js request-error reporting, bounded flushes, and shutdown drains.
- Wire PostHog error tracking into gateway, platform, proxy, shell, www, and first-party social routes while keeping expected 4xx responses and disabled-env paths quiet.
- Preserve thrown Hono HTTPException responses, including 5xx status/body and Hono-like errors from alternate module instances, while still capturing 5xx exceptions to PostHog.
- Keep first-party social route handlers aligned with Hono HTTPException response semantics and restore fallback Hono error logging for generic exceptions.
- Keep shell browser telemetry from using relative `/ingest` API hosts it cannot proxy, while preserving www `/ingest` support.
- Propagate PostHog env/build args through Docker, compose, host-bundle, user-image, and provisioned-container flows without passing PostHog secrets to external Conduit containers.
- Address review feedback for nonblocking telemetry, built package exports, Next unregister hooks, shell client routing, Docker dev ownership, public build args, proxy shutdown ordering, effective proxy forced-exit behavior, documented PostHog SDK timeout limitations, and parallel platform PostHog shutdown drains.

## Tests
- bun run test tests/observability/posthog-error-tracking.test.ts tests/gateway/social-api.test.ts tests/platform/social-api.test.ts
- bun run test tests/observability/posthog-error-tracking.test.ts
- bun run typecheck
- bun run check:patterns
- NEXT_PUBLIC_CLERK_PUBLISHABLE_KEY=pk_test_Y2xlcmsuZXhhbXBsZS5jb20k pnpm --filter './shell' build
- NEXT_PUBLIC_CLERK_PUBLISHABLE_KEY=pk_test_Y2xlcmsuZXhhbXBsZS5jb20k pnpm --filter './www' build
- bun run test (latest local attempt failed due ENOSPC on the temp filesystem with 129 MiB free; GitHub CI full suite is required after latest push)

## Invariants
- Source of truth: PostHog is configured only from service environment variables; app data and auth state remain in their existing stores.
- Lock/transaction scope: no database transaction is introduced; telemetry capture is outside user-visible response paths and is bounded by flush/shutdown timeouts. PostHog SDK requests cannot be aborted by the SDK, so fire-and-forget capture keeps request handlers from waiting on them.
- Acceptable orphan states: when PostHog env is absent, capture fails, or a flush times out, the app continues and telemetry may be dropped.
- Auth source of truth: existing gateway/platform/shell auth remains authoritative; PostHog distinct IDs are telemetry-only and sanitized request metadata strips secrets and query values.
- Deferred scope: this PR adds error tracking only, not product analytics events, session replay policy changes, or instrumentation for the external Conduit image.


HamedMP:
Addressed the latest Greptile quality notes in commit 41f5b04: added a code comment documenting that the PostHog SDK timeout bounds only our await because the SDK exposes no AbortSignal, and added a regression/static check for that. For the suggested isShutDown guard in www/src/lib/posthog-server.ts, I intentionally did not disable warm-process reinitialization: shutdownPostHog() is also used as an action-level flush by dashboard/Inngest flows, and permanently returning the noop client afterward would drop legitimate later telemetry in the same warm process. I documented that behavior and covered it in the static check.

PR #94, opened 2026-05-12
fix(gateway): use function-based watcher ignore to prevent event loop starvation

ArianMoeini:
## Summary
- Replace glob-based chokidar `ignored` patterns with a segment-matching function
- Glob patterns like `**/node_modules/**` only match paths *inside* directories, not the directory entries themselves — chokidar still calls `readdir` on them before filtering contents
- On VPSes where `MATRIX_HOME` contains large trees (e.g. `projects/matrix-os-dev/node_modules` at 2.3GB), the initial poll scan overwhelms the event loop and prevents the gateway from binding its port

## Root cause
The gateway creates a chokidar file watcher with `usePolling: true` on the entire `MATRIX_HOME`. Even with `ignoreInitial: true`, chokidar walks the directory tree to establish a baseline for change detection. The `**/node_modules/**` glob matches `node_modules/express/index.js` but NOT `node_modules` itself — so chokidar recurses into it, runs `readdir` on thousands of entries, and saturates the event loop at 200%+ CPU while growing to 6GB RSS.

## Fix
A function-based `ignored` option that checks path segments catches directories before `readdir` runs. Gateway startup drops from infinite hang to ~5 seconds.

## Test plan
- [x] `bun run test tests/gateway/watcher.test.ts` — 4 tests pass
- [x] Verified gateway starts and binds port 4000 on this VPS with the function-based ignore
- [x] Verified gateway hangs indefinitely with the old glob-based ignore


HamedMP:
Updated this branch against current main and kept the watcher ignore salvage. Also tightened the matrix.db predicate to only match the final path segment, addressing the Greptile concern about accidentally ignoring matrix.db-looking directory names.

PR #128, opened 2026-05-14
fix(symphony): start Codex sessions through zellij

HamedMP:
## Summary

- Fix Linear ticket preview GraphQL variables so Linear IDs use `ID!` while state/label filters remain strings.
- Generate valid zellij KDL pane properties and start zellij sessions through a retained node-pty process instead of execFile.
- Launch Codex in non-interactive `codex exec` mode with repo trust bypass and approval disabled, and create a safe default `WORKFLOW.md` for registered projects when missing.

## Tests

- `pnpm exec vitest run tests/gateway/symphony-linear-source.test.ts tests/gateway/zellij-runtime.test.ts tests/gateway/agent-launcher.test.ts tests/gateway/symphony-workflow.test.ts tests/gateway/symphony-orchestrator.test.ts`
- `bun run typecheck`
- `bun run check:patterns` (0 violations; existing scanner warnings only)
- `bun run test` was started after the focused suite and typecheck, but produced no worker output for several minutes and was stopped with SIGTERM to avoid leaving a hung local process.

## Invariants

- Source of truth: Symphony installation/run state remains in the existing Symphony repository; project workflow contracts live in the registered project root as `WORKFLOW.md`, with a generated default only when the default path is missing.
- Lock/transaction scope: this change does not add database writes or change repository transaction boundaries; dispatch still claims the run before creating the worktree and session.
- Process lifecycle: zellij startup now requires a PTY and is considered running only after the startup delay without exit; retained PTYs are capped and killed before `zellij kill-session` on cleanup.
- Acceptable orphan states: if worktree creation or runtime startup fails, existing dispatch failure handling keeps the run in retrying/blocked with generic error codes and preserves worktree metadata when available.
- Auth source of truth: Linear credentials remain server-side, and Codex launches in workspace-write sandbox mode unless the existing explicit admin override is provided.
- Deferred scope: this PR does not change dashboard UI, Linear state transitions, or production deployment; it only fixes preview/query, workflow preflight, launch args, and zellij runtime startup.

### DIFF
--- a/packages/gateway/src/agent-launcher.ts
+++ b/packages/gateway/src/agent-launcher.ts
@@ -99,7 +99,14 @@ export function buildAgentLaunch(input: AgentLaunchInput): AgentLaunchSpec {
     case "codex":
       return {
         command,
-        args: [...codexSandboxArgs(input.sandbox), ...promptArgs(input.prompt)],
+        args: [
+          "exec",
+          "--skip-git-repo-check",
+          "--ask-for-approval",
+          "never",
+          ...codexSandboxArgs(input.sandbox),
+          ...promptArgs(input.prompt),
+        ],
         cwd: input.cwd,
         env: {},
       };
--- a/packages/gateway/src/symphony/linear-source.ts
+++ b/packages/gateway/src/symphony/linear-source.ts
@@ -138,10 +138,10 @@ function includesAllLabels(ticket: TrackedTicket, requiredLabels: string[]): boo
 }
 
 function buildIssuesQuery(input: { projectId?: string; state?: string; labelName?: string; assigneeId?: string }) {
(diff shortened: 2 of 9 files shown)


HamedMP:
@greptile review

Please rerun on current head `b20a3bc6`. The previous 3/5 review comments have been addressed: Codex exec restores `--` before prompt args, retained zellij PTYs now have TTL eviction, and zellij PTY env now uses a safe process-env allowlist plus explicit launch env.

HamedMP:
@greptile review

Please rerun on current head `6188d5a7`. Follow-ups from the 5/5 summary are addressed too: `codexPromptArgs` was removed in favor of the shared `promptArgs`, and zellij retains the PTY immediately after spawn so `kill()` can reach startup-in-progress sessions before the delay resolves.

PR #147, opened 2026-05-17
fix(canvas): preserve app state across fullscreen transitions

ArianMoeini:
## Summary
- **Root cause**: In canvas mode, entering fullscreen created a brand-new `<AppViewer>` (and iframe) as an overlay. When exiting fullscreen, this overlay was destroyed and the canvas showed the original (stale) iframe — losing all in-memory state.
- **Fix**: The existing `CanvasWindow` now renders fullscreen in-place using inverse canvas transform positioning (`scale(1/zoom)` counter-transform), keeping the same React tree and iframe mounted throughout the transition.
- Removed the canvas-specific fullscreen overlay from `Desktop.tsx` that was creating duplicate component instances.

## How it works
When a canvas window goes fullscreen:
1. `useLayoutEffect` walks ancestor DOM nodes, setting `overflow: visible` and `z-index: 100` so the window can extend beyond the canvas container to cover the full viewport (including dock).
2. The window's position is computed as the inverse of the canvas transform: `left: -contLeft/zoom - panX`, `top: -contTop/zoom - panY`, with `transform: scale(1/zoom)` to counter the parent's `scale(zoom)`.
3. Title bar, resize handle, and click-to-interact overlays are hidden.
4. On exit, all ancestor styles are restored to their original values.

## Test plan
- [ ] Open any app (e.g. Notes) in canvas mode, type some content, enter fullscreen — content should still be there
- [ ] Exit fullscreen — content should still be there (not lost)
- [ ] Verify non-canvas mode fullscreen still works (was not affected by this change)
- [ ] Verify fullscreen exit pill (top-left hover zone) renders above the fullscreen window
- [ ] Verify Escape key exits fullscreen
- [ ] Test with different canvas zoom levels
- [ ] Test window resize while in fullscreen


HamedMP:
Closing this older duplicate fullscreen-state PR in favor of keeping #148 as the salvage candidate. #148 is smaller/newer, and current main already contains the core in-place fullscreen preservation work.

PR #148, opened 2026-05-17
fix(canvas): preserve app state across fullscreen transitions

ArianMoeini:
## Summary
- **Root cause**: In canvas mode, entering fullscreen created a brand-new `<AppViewer>` (and iframe) as an overlay in `Desktop.tsx`. When exiting, that overlay was destroyed and the canvas showed the original stale iframe — losing all in-memory state (typed text, scroll position, etc.).
- **Fix**: The existing `CanvasWindow` now renders fullscreen in-place using inverse canvas-transform positioning (`scale(1/zoom)` counter-transform), keeping the same React tree and iframe mounted throughout the transition. No duplicate components created.
- Removed the canvas-specific fullscreen overlay from `Desktop.tsx`.

## How it works
When a canvas window goes fullscreen:
1. `useLayoutEffect` walks ancestor DOM nodes, setting `overflow: visible` and `z-index: 100` so the window can extend beyond the canvas container to cover the full viewport (including dock).
2. The window's position is computed as the inverse of the canvas transform: `left: -contLeft/zoom - panX`, `top: -contTop/zoom - panY`, with `transform: scale(1/zoom)` to counter the parent's `scale(zoom)`.
3. Title bar, resize handle, and click-to-interact overlays are hidden.
4. On exit, all ancestor styles are restored to their original values.

## Verified with Playwright
Tested the full enter → during → exit fullscreen cycle on a live Terminal canvas window:
- Terminal content (`[Session expired...] matrix@matrix-arian:~/projects$`) preserved identically through all three phases
- No old-style duplicate overlay created (`oldOverlayCount: 0`)
- Same number of canvas windows (3) and iframes (1) throughout — no duplication
- Fullscreen fills viewport exactly (wrapper rect matches viewport size)
- Exit restores normal canvas positioning and card styling

## Test plan
- [x] Terminal in canvas mode: state preserved through fullscreen cycle (Playwright verified)
- [ ] Verify non-canvas (floating windows) mode fullscreen still works (not affected — different code path)
- [ ] Test with iframe apps (Notes, etc.) when gateway is connected
- [ ] Verify fullscreen exit pill renders above the fullscreen window
- [ ] Test with different canvas zoom levels
- [ ] Test window resize while in fullscreen


ArianMoeini:
Addressed both Codex review comments in 4b0f9cb9:

1. **Canvas hit testing** — `data-canvas-window` attribute is now removed when fullscreen (`!isPreview && !isFullscreen`), so `CanvasTransform` won't treat fullscreen interactions as canvas-surface events that clear focus and enable panning.

2. **Mobile prop** — Added `useMobileViewport()` hook and now pass `mobile={isMobile}` to `TerminalApp`, `FileBrowser`, and `ChatApp` in the canvas fullscreen path, matching what the removed Desktop overlay did.

ArianMoeini:
Addressed Greptile review in b224b7b9:

**P2 — container offset fallback**: Added fallback to `el.parentElement.parentElement` (the transform div's parent) when no `overflow:hidden` ancestor is found. This prevents the fullscreen position from silently defaulting to `(0, 0)` if clipping uses `clip-path` or another mechanism.

**P1 — `iconRegenerationInProgress` flag**: This is a pre-existing issue in `packages/gateway/src/server.ts`, not part of this PR's changes (which only touch `CanvasWindow.tsx` and `Desktop.tsx`). The gateway files appear in the diff because the branch base diverged. Happy to fix in a separate PR if needed.

ArianMoeini:
Addressed remaining Codex comment in 98c27fd3:

**P2 — Pass onOpenApp to AppViewer**: Added `onOpenApp={(name, path) => openWindow(name, path, 0)}` to the `<AppViewer>` in the canvas render path, so apps can launch other apps via the `os:open-app` bridge message during fullscreen (matching what the removed Desktop overlay did).

All automated review comments are now addressed across 4 commits.

HamedMP:

Closing as superseded by the implementation already on `main`. The current `CanvasWindow` keeps the same app tree mounted during fullscreen and uses the inverse canvas transform; the review fixes from [4b0f9cb](https://github.com/HamedMP/matrix-os/commit/4b0f9cb9eb) and [b224b7b](https://github.com/HamedMP/matrix-os/commit/b224b7b9ad) are also present. Reviving this stale branch would reintroduce an older `Desktop`/Canvas shape without adding missing behavior.

PR #230, opened 2026-05-28
fix(platform): tolerate older VPS bundles without Symphony

HamedMP:
## Summary
- Treat the Symphony host binary as optional during customer VPS cloud-init bootstrap.
- Enable/start `matrix-symphony.service` only when both the binary and unit exist in the extracted host bundle.
- Keep core restore/gateway/shell/sync services fail-closed and required.

## Incident Context
Nima's recovery VM booted an older bundle (`v2026.05.24-082-staging-ff593047`) that did not include `/opt/matrix/bin/matrix-symphony`. Current cloud-init treated that binary as required, so setup aborted before gateway startup and platform registration.

## Invariants
- Source of truth: platform `user_machines` remains canonical for VPS state; this only changes bootstrap service selection.
- Lock/transaction scope: no database writes or external platform state changes in this PR.
- Acceptable orphan states: older bundles without Symphony still boot core services; Symphony remains disabled until a bundle ships the binary and unit.
- Auth source of truth: unchanged; gateway/platform bearer token checks are preserved.
- Deferred scope: does not change stale recovery reconciliation or registration-token retry behavior.

## Verification
- `bun run test tests/deploy/customer-vps/symphony-systemd.test.ts`


HamedMP:

Status: conflicts with `main`. Customer bundles now ship and condition the current Elixir Symphony runtime differently, and cloud-init/systemd coverage has expanded well beyond this branch.

To revive: reproduce the old-bundle bootstrap case against the current immutable host-bundle contract, decide whether compatibility with bundles missing Symphony is still supported, then add a focused current-template test before changing enable/start behavior. Owner keep/kill decision requested.

PR #232, opened 2026-05-28
fix(proxy): default DB path to home dir and auto-create parents

maciej-konczal:
Running bun run dev I got: "  TypeError: Cannot open database because the directory does not exist
    at getDb (packages/proxy/src/db.ts:9:10)"

### DIFF
--- a/packages/proxy/package.json
+++ b/packages/proxy/package.json
@@ -4,7 +4,7 @@
   "private": true,
   "type": "module",
   "scripts": {
-    "dev": "tsx watch src/main.ts",
+    "dev": "tsx watch --env-file-if-exists=../../.env src/main.ts",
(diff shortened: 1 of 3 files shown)


HamedMP:
thanks for the pr @maciej-konczal . we have moved away from sqlite, to postgres only. sqlite has been the old db and soem coding agents (mainly claude) still makes mistake. we should move completely to codex. 

The proper fix is to migrate the proxy persistence tables (api_usage, user_quotas) to Postgres/Kysely using DATABASE_URL or the existing platform/owner Postgres wiring, then remove better-sqlite3 from @matrix-os/proxy.

maciej-konczal:
Closing and https://github.com/HamedMP/matrix-os/pull/235 provided

PR #233, opened 2026-05-28
fix(onboarding): notify user when voice intro falls back due to

maciej-konczal:
Gemini live does not on onboarding without Gemini API Key


HamedMP:

Status: conflicts with `main`; onboarding voice, readiness, and shell error presentation have been rewritten. Current code falls back from unavailable voice to text, but this branch's user-notice contract has not been carried forward as-is.

To revive: define the current user-facing fallback copy, route it through the shared shell notification host, add gateway/hook/component tests, and avoid exposing provider details. Owner keep/kill decision requested.

PR #235, opened 2026-05-28
refactor(proxy): migrate from SQLite to Postgres via Kysely

maciej-konczal:
Replaces better-sqlite3 with Kysely + pg, matching the project-wide "PostgreSQL via Kysely" rule in CLAUDE.md. Adds PROXY_DATABASE_URL with fallback to PLATFORM_DATABASE_URL, removes the SQLite file path default and the proxy-data Docker volume. Tests use kysely-pglite (in-memory PG) following tests/platform/platform-db-test-helper.ts.

Behavior change: host-mode bun run dev now requires a Postgres available (Homebrew or Docker). README documents both paths. Docker stack unchanged for users -- PROXY_DATABASE_URL defaults to the existing postgres service.

Also wires DB pool cleanup into the SIGTERM/SIGINT shutdown handler via resetProxyDb() so the pool doesn't outlive graceful shutdown.

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -11,6 +11,10 @@ GEMINI_API_KEY=
 PORT=4000
 MATRIX_HOME=
 
+# Proxy (api usage + quotas)
+# Defaults to PLATFORM_DATABASE_URL when unset. Postgres only (Kysely).
+PROXY_DATABASE_URL=
+
 # Shell (Next.js)
 NEXT_PUBLIC_GATEWAY_WS=ws://localhost:4000/ws
 NEXT_PUBLIC_GATEWAY_URL=http://localhost:4000
--- a/README.md
+++ b/README.md
@@ -224,6 +224,29 @@ bun run test:integration  # Integration tests (needs API key, uses haiku)
 bun run test:coverage     # Coverage report
 ```
 
+### Postgres (host dev only)
+
+Matrix OS uses PostgreSQL for all persistence (platform, kernel durable state, proxy usage tracking). The Docker dev stack runs Postgres for you; if you're running on the host via `bun run dev`, you need a Postgres available.
+
+The easiest path on macOS:
+
+```bash
+brew install postgresql@17           # one-time
+brew services start postgresql@17
+createdb matrixos_platform
+```
+
+Then in your `.env`:
+
+```bash
+PROXY_DATABASE_URL=postgresql://localhost:5432/matrixos_platform
+PLATFORM_DATABASE_URL=postgresql://localhost:5432/matrixos_platform
+```
+
+Homebrew's Postgres uses your local user with no password (local trust auth). The proxy creates its tables automatically on first boot.
+
+Don't need the proxy locally? Skip Postgres entirely and run only the gateway + shell: `bun run dev:gateway & bun run dev:shell`.
+
 ### Start Development
 
 ```bash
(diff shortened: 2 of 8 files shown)


HamedMP:
I would hold this PR until the following are addressed.

Required before merge:

1. Fix the red CI shard. `Unit Tests (3/4)` is still failing on job `78523030952` because `tests/observability/posthog-error-tracking.test.ts` still source-inspects for the exact string `await posthogErrorTracker.shutdown()`, while `packages/proxy/src/main.ts` now calls it inside `Promise.allSettled([...])`. This needs a test update that still verifies shutdown ordering and drain behavior.

2. Update the distro compose proxy services, not only `docker-compose.dev.yml`. On the current head, these still pass only the removed SQLite env/volume path:
   - `distro/docker-compose.local.yml`: `PROXY_DB_PATH=/data/proxy.db`
   - `distro/docker-compose.multi.yml`: `PROXY_DB_PATH=/data/proxy.db`
   - `distro/docker-compose.platform.yml`: `PROXY_DB_PATH=/data/proxy.db`

   Since `packages/proxy/src/db.ts` now throws without `PROXY_DATABASE_URL` or `PLATFORM_DATABASE_URL`, those proxy services will fail startup/health. Mirror the dev compose pattern: add a Postgres dependency, set `PROXY_DATABASE_URL` to the platform Postgres database, and remove the obsolete `proxy-data` volume if it is no longer used.

3. Platform compose needs explicit consideration. `distro/docker-compose.platform.yml` already has `postgres` and gives `PLATFORM_DATABASE_URL` to the platform service, but the proxy service does not receive that URL. Because platform-created user containers default to `PROXY_URL=http://proxy:8080` and `ANTHROPIC_BASE_URL=http://proxy:8080`, a dead proxy breaks platform-created runtimes, not just proxy metrics.

4. Add the mandatory backend PR body `Invariants` section:
   - Source of truth
   - Lock/transaction scope
   - Acceptable orphan states
   - Auth source of truth
   - Deferred scope

5. Resolve or explicitly defer the two Greptile findings in the PR body with a linked follow-up if not fixed here: transactional schema migration and the `POSTGRES_URL` fallback appending `/matrixos_platform` to a URL that may already include a database name.

Related platform wiring check, possibly pre-existing but worth handling while touching platform compose: the platform orchestrator generates `sk-proxy-*` API keys from `PLATFORM_SECRET`, while proxy validation only works if the proxy receives `PROXY_SHARED_SECRET` or `PLATFORM_SECRET`. `distro/docker-compose.platform.yml` currently passes `PLATFORM_SECRET` to `platform` but not to `proxy`. If this PR updates platform proxy env, please verify the proxy shared-secret/admin-token path end-to-end, including platform calls to `/usage/summary` and user container calls through `ANTHROPIC_BASE_URL=http://proxy:8080`.

Current merge status I see: head `39a89e0e`, Greptile reports 5/5 but with the two noted findings, GitHub CI has one failing unit shard, and Vercel is blocked on deployment authorization.

HamedMP:

Status: conflicts with `main`, but the architectural goal still matches the PostgreSQL/Kysely constitution. Proxy database wiring, shutdown ownership, Docker/dev configuration, and lockfile state have moved since this head.

To revive: rebuild from current `packages/proxy`, use an owner-managed Kysely/Postgres resource with explicit shutdown ownership, replace stale test/dependency assumptions, run `pnpm install` at the repo root, and land with migration/rollback coverage. Owner keep/kill decision requested.

PR #282, opened 2026-05-30
fix(platform): pass customer VPS image channel env

HamedMP:
## Summary
- Passes CUSTOMER_VPS_IMAGE_VERSION through the production platform compose service so the provisioner can use the stable channel for new customer VPSes.

## Tests
- Live production platform recreated with CUSTOMER_VPS_IMAGE_VERSION=stable and /health returns ok.
- Existing fleet release probe reports all running VPSes on v2026.05.30-stripe-stable-e9093ba0b.

## Invariants
- Source of truth: platform env remains the source of the customer VPS default image/channel; compose now passes it into the platform process.
- Lock/transaction scope: configuration-only change; no database writes.
- Acceptable orphan states: existing VPSes keep their installed bundle; this affects future provisioning defaults only.
- Auth source of truth: unchanged PLATFORM_SECRET/Clerk auth handling.
- Deferred scope: production Stripe live keys and price IDs remain operator-provided env setup.


HamedMP:
[codex] Addressed Greptile's stable/default mismatch finding by changing the application fallback in customer-vps-config.ts to stable and updating the provisioning tests to assert stable channel pinning.\n\nCommit: adb240144\nTests: pnpm exec vitest run tests/platform/customer-vps.test.ts; bun run check:patterns:diff

PR #287, opened 2026-05-31
fix(billing): serve unpaid shell and avoid customer write

HamedMP:
## Summary
- Allow unpaid signed-in users to load the shell document and static assets so the locked Billing settings page can render instead of raw 402 JSON.
- Keep runtime data paths, apps, files, modules, APIs, code-domain proxying, and websockets gated until billing access is active.
- Stop pre-creating Stripe Customers before Checkout. Checkout now creates the customer when needed, and signed subscription webhooks link the customer back through server-written Clerk metadata.

## Tests
- `bun run test tests/platform/proxy-routing.test.ts`
- `bun run test tests/platform/billing-routes.test.ts tests/platform/stripe-billing.test.ts`
- `pnpm --filter '@matrix-os/platform' exec tsc --noEmit -p tsconfig.typecheck.json`
- `bun run check:patterns:diff` (warnings only; no violations)

## Review/Monitoring
- Hotfixed production platform image from this branch after the restricted live Stripe key failed customer creation but succeeded Checkout Sessions.
- Needs CI and Greptile monitoring before merge.

## Invariants
- Source of truth: Stripe subscription webhooks plus `billing_customers`/`billing_entitlements` remain canonical for paid runtime access; internal test/operator access stays in `billing_entitlement_overrides`.
- Lock/transaction scope: webhook event insertion, customer link creation from signed metadata, and entitlement upsert run inside the existing billing webhook transaction; Checkout performs no entitlement writes.
- Acceptable orphan states: a Checkout Session may exist before webhook delivery; until the subscription webhook lands, runtime access remains inactive and the user stays on the billing surface.
- Auth source of truth: Clerk-authenticated platform requests can load the shell paywall; runtime data/API/websocket access still requires active entitlement decision.
- Deferred scope: full self-serve plan switching/add-on reconciliation remains outside this hotfix.

### DIFF
--- a/packages/platform/src/billing-routes.ts
+++ b/packages/platform/src/billing-routes.ts
@@ -9,7 +9,6 @@ import {
   insertBillingWebhookEvent,
   runBillingWebhookTransaction,
   upsertBillingEntitlement,
-  type BillingCustomerRecord,
   type PlatformDB,
 } from './db.js';
 import {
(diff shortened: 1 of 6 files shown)


HamedMP:
@greptileai

PR #294, opened 2026-05-31
fix(platform): bridge signed-in Clerk auth sessions

HamedMP:
## Summary
- exchange browser-only Clerk sessions for an HttpOnly Matrix app session cookie before redirecting to the shell
- route the app-session endpoint on the platform instead of proxying it to a user runtime
- replace the signed-in auth dead-end with loading, retry, sign-out, and no-runtime states

## Invariants
- Source of truth: Clerk remains the identity source; platform user_machines/containers remain the runtime routing source.
- Lock/transaction scope: no multi-write transaction is introduced; the endpoint only verifies Clerk auth, reads runtime state, and issues a signed cookie.
- Acceptable orphan states: if the user has no active runtime, no app cookie is issued and the client shows a no-runtime state.
- Auth source of truth: Clerk bearer token is verified server-side before a short-lived Matrix app session JWT is set as HttpOnly/SameSite=Lax.
- Deferred scope: this does not auto-provision or remap existing VPS ownership for newly recreated Clerk users.

## Validation
- bun run test tests/platform/proxy-routing.test.ts -t "app session|signed-in auth pages|no runtime"
- bun run test tests/platform/proxy-routing.test.ts
- bun run typecheck
- bun run check:patterns

### DIFF
--- a/packages/platform/src/main.ts
+++ b/packages/platform/src/main.ts
@@ -105,6 +105,7 @@ const VPS_RUNTIME_METRICS_TTL_MS = 45_000;
 const DOCKER_INSPECT_TIMEOUT_MS = 10_000;
 const CODE_SERVER_PORT = Number(process.env.MATRIX_CODE_SERVER_PORT ?? 8787);
 const CODE_SESSION_COOKIE = 'matrix_code_session';
+const APP_SESSION_COOKIE = 'matrix_app_session';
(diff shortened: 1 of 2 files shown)


HamedMP:
@greptileai review current head 575b7003ac60a64c61fb45f21bf11d78907545ad

PR #349, opened 2026-06-03
fix(apps): route sandboxed app assets with signed urls

HamedMP:
## Summary
- keep app-session cookies `SameSite=Strict` and platform route cookies `SameSite=Lax`
- let the gateway serve read-only Vite asset files (`GET`/`HEAD /apps/{slug}/assets/*`) without an app-session cookie while keeping app HTML and mutating requests session-gated
- rewrite proxied Vite app HTML so `srcdoc` iframes load assets from explicit `/vm/{handle}/apps/{slug}/assets/*` URLs that do not depend on null-origin cookies
- allow explicit VM Vite asset routes to proxy with null-origin CORS without browser cookies

## Why
Default Vite apps render from a sandboxed `srcdoc` iframe with origin `null`. Browser cookie SameSite rules prevent the iframe from sending the route/session cookies on module script and stylesheet loads, so assets return 401 and the app opens white. Relaxing auth cookies to `SameSite=None` would fix the symptom but broaden CSRF/null-origin exposure. This keeps cookies protected and removes the cookie dependency only for static built asset files.

## Invariants
- Source of truth: app access remains the gateway-signed `matrix_app_session__{slug}` cookie for app HTML, APIs, and mutating app paths; runtime selection remains platform machine records plus explicit `/vm/{handle}` routing.
- Lock/transaction scope: no database writes or transactions changed.
- Acceptable orphan states: already-rendered old srcdoc HTML may still reference `/apps/{slug}/assets/*`; re-opening or refreshing the app fetches rewritten HTML. Static built assets contain app code, not owner app data.
- Auth source of truth: app data/API/mutation access remains app-session protected; only read-only built Vite assets bypass the app-session middleware. Explicit VM asset routing does not attach a platform user identity when no user is authenticated.
- Deferred scope: this does not change app sandbox CSP, MatrixOS bridge permissions, app data APIs, or release-channel UI.

## Validation
- bun run test tests/gateway/app-runtime/app-session.test.ts tests/gateway/app-runtime/app-session-middleware.test.ts tests/platform/proxy-routing.test.ts
- bun run typecheck
- bun run check:patterns
- git diff --check

### DIFF
--- a/packages/gateway/src/app-runtime/app-session-middleware.ts
+++ b/packages/gateway/src/app-runtime/app-session-middleware.ts
@@ -1,7 +1,7 @@
 import { readFileSync } from "node:fs";
-import { timingSafeEqual } from "node:crypto";
 import { join } from "node:path";
 import type { MiddlewareHandler } from "hono";
+import { timingSafeStringEquals } from "../security/timing-safe.js";
 import { verifyAppSession, type AppSessionPayloadType } from "./app-session.js";
 import { SAFE_SLUG } from "./manifest-schema.js";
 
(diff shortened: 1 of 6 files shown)


HamedMP:
@greptileai please re-review the latest head 03b7eddcf3cd8875f240cd51655632b16a7b7d0b. The prior 4/5 summary is stale and its listed issues were addressed in the latest commit.

PR #371, opened 2026-06-05
fix(gateway): start zellij shell sessions through pty

Nima-Naderi:
## Summary

- Replace headless zellij session creation with a retained PTY startup path so `matrix run -it` does not fail on `attach --create-background` timeouts.
- Keep existing zellij attach/action behavior while adding bounded cleanup for retained creation PTYs.
- Preserve numeric PTY startup-exit signals in sanitized diagnostics and close the retained-PTY exit-listener ordering window.
- Update gateway tests for PTY startup, startup failures, delete cleanup, retained PTY cap eviction, and numeric signal diagnostics.

## Tests

- `pnpm test tests/gateway/shell-zellij.test.ts`
- `pnpm test tests/cli/run.test.ts`
- `pnpm --filter '@matrix-os/gateway' exec tsc --noEmit`
- `pnpm run check:patterns`
- `git diff --check`

## Review/Monitoring

- Greptile initially reported `4/5` on commit `d48d4e64` for dropped numeric PTY signal diagnostics.
- Fixed in commit `cdca9c0a`; awaiting fresh Greptile review on the latest commit.
- CI checks are being monitored from this PR branch.

## Invariants

- **Source of truth**: zellij remains the canonical source for shell session runtime state; Matrix metadata continues to reference sessions by name.
- **Lock/transaction scope**: no database transaction is involved. Retained creation PTYs are bounded in an adapter-local map with TTL sweep, cap eviction, and explicit delete-time cleanup.
- **Acceptable orphan states**: if zellij starts but the initial PTY cannot be retained, later attach/delete paths still target the named zellij session. Retained PTYs are best-effort helpers and are evicted by TTL/cap/delete.
- **Auth source of truth**: unchanged. Shell routes and WebSocket attach paths continue to enforce their existing gateway auth before reaching this adapter.
- **Deferred scope**: this does not implement non-interactive `matrix run --` execution or change CLI auth/profile behavior.

### DIFF
--- a/packages/gateway/src/shell/zellij.ts
+++ b/packages/gateway/src/shell/zellij.ts
@@ -308,22 +308,24 @@ export function createZellijAdapter(deps: ZellijAdapterDeps = {}): ZellijAdapter
         throw safeZellijError(err);
       }
       const startup = { exited: null as PtyExitEvent | null };
+      const retained: RetainedCreatePty = {
+        process: pty,
+        startedAtMs: nowMs(),
+        exitDisposable: null,
+      };
+      retainedCreatePtys.set(options.name, retained);
       const exitDisposable = pty.onExit((event) => {
         startup.exited = event;
         releaseRetainedCreatePty(options.name);
       });
-      retainedCreatePtys.set(options.name, {
-        process: pty,
-        startedAtMs: nowMs(),
-        exitDisposable,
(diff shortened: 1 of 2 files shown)


Nima-Naderi:
@greptileai please review this man asap :)

PR #373, opened 2026-06-05
feat(platform): add Cloud Run image workflow

HamedMP:
## Summary

- Add a platform-only Dockerfile for the Cloud Run control plane.
- Add a Cloud Build config that builds Dockerfile.platform instead of the customer/runtime image.
- Add a workflow-dispatch Cloud Run deploy workflow with tagged candidate smoke and optional manual promotion.

## Invariants

- Source of truth: platform Postgres remains canonical for platform records; Artifact Registry stores deployable image bytes only.
- Lock/transaction scope: no database transaction behavior changes; this PR only changes build/deploy plumbing.
- Acceptable orphan states: an image can be built without being promoted; a candidate revision can exist with zero traffic until smoke/promotion completes.
- Auth source of truth: Cloud Run runtime auth remains Clerk, PLATFORM_SECRET, and per-route checks. GitHub deployment auth uses GCP workload identity through environment secrets.
- Deferred scope: does not point production DNS to Cloud Run, does not migrate Neon data, does not add the missing Inngest/Pipedream webhook secret versions, and does not fan out deploy to VPSes.
- Rollback plan: keep the previous Cloud Run revision or route DNS back to the legacy platform; the workflow supports candidate revisions before promotion.
- PostHog events/errors: unchanged; this is deployment-image and workflow plumbing only.

## Verification commands

- git diff --check
- Manual smoke used equivalent temporary Dockerfile.platform image: Cloud Run /health returned {"status":"ok"} on matrix-platform-00003-ros

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -2,6 +2,7 @@ node_modules
 .git
 .gitignore
 Dockerfile*
+!Dockerfile.platform
 .dockerignore
 *.log*
 .env*
--- a/.github/workflows/platform-cloud-run.yml
+++ b/.github/workflows/platform-cloud-run.yml
@@ -1,0 +1,157 @@
+name: Platform Cloud Run
+
+on:
+  workflow_dispatch:
+    inputs:
+      environment:
+        description: GitHub environment to deploy
+        required: true
+        default: staging
+        type: choice
+        options:
+          - staging
+          - production
+      promote:
+        description: Promote the deployed revision to 100% traffic
+        required: true
+        default: false
+        type: boolean
+
+permissions:
+  contents: read
+  id-token: write
+
+concurrency:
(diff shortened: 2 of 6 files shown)


HamedMP:
[codex] Addressed Greptile findings in 56ee47eab:\n- Added non-root runtime user for Dockerfile.platform.\n- Preserved the existing .dockerignore secret exclusions and added an explicit Dockerfile.platform exception.\n- Converted the platform image to a two-stage build so compiler/native build tools stay out of runtime.\n- Removed production fallbacks from deploy workflow and added required-var validation.\n- Made CUSTOMER_VPS_IMAGE_VERSION environment-specific via GitHub environment vars.

PR #387, opened 2026-06-06
fix(platform): fail closed on host bundle signer storage

HamedMP:
## Summary
- fail closed when customer VPS mode lacks a dedicated host-bundle R2 store instead of falling back to the sync bucket
- reject bundle storage config that points at the same bucket as sync storage
- extend Platform Cloud Run candidate smoke to verify a signed dev host-bundle manifest and one-byte bundle read before promotion

## Invariants
- Source of truth: platform Postgres remains the release metadata and channel source of truth; R2 only stores immutable bundle bytes.
- Lock/transaction scope: no database writes are added; startup config validation runs before opening the platform DB, and route reads continue through existing release lookup helpers.
- Acceptable orphan states: a misconfigured platform revision may fail startup or return host-bundle storage unavailable; it must not sign bundle URLs from the sync bucket.
- Auth source of truth: public host-bundle metadata/download routes remain public; release registration and channel promotion still require the platform admin bearer token.
- Deferred scope: Neon database migration and production domain cutover are not included in this PR.

## Verification
- bun run test tests/platform/home-mirror-env-check.test.ts tests/platform/host-bundle-route.test.ts tests/platform/customer-vps-host-bundle.test.ts
- bun run check:patterns
- bun run typecheck

### DIFF
--- a/.github/workflows/platform-cloud-run.yml
+++ b/.github/workflows/platform-cloud-run.yml
@@ -206,8 +206,22 @@ jobs:
             echo "Candidate host bundle channel smoke did not return a signed bundle URL." >&2
             exit 1
           fi
-          if printf '%s\n' "$bundle_url" | grep -q 'matrixos-sync'; then
-            echo "Candidate host bundle signer returned the sync bucket; refusing to promote." >&2
+          sync_bucket="$(gcloud secrets versions access latest --secret=r2-bucket)"
+          bundle_bucket="$(gcloud secrets versions access latest --secret=r2-bundles-bucket)"
+          if [ -z "$sync_bucket" ] || [ -z "$bundle_bucket" ]; then
+            echo "Candidate host bundle smoke could not read required bucket secret names." >&2
+            exit 1
+          fi
+          if [ "$bundle_bucket" = "$sync_bucket" ]; then
+            echo "Dedicated host bundle bucket secret matches the sync bucket secret; refusing to promote." >&2
+            exit 1
+          fi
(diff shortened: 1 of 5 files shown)


HamedMP:
[codex] Accepting Greptile's three findings. I will:
- make the Cloud Run smoke compare the actual GCP Secret Manager sync and bundle bucket names, and keep the URL string check as a native-R2-url defense without treating it as the only guard;
- pull the default sync bucket into a named platform constant;
- add a route-level dev-channel regression with CUSTOMER_VPS_ENABLED=true proving the channel manifest uses hostBundleObjectStore, not customerVpsObjectStore.

I will push a follow-up commit and let CI/Greptile rerun.

HamedMP:
[codex] Changes since the 4/5 Greptile review:
- Accepted the workflow smoke concern: the candidate smoke now reads the configured sync and bundle bucket secret names, rejects matching buckets, rejects signed URLs containing the configured sync bucket, and rejects native R2 URLs outside the configured bundle bucket before doing the byte-range read.
- Accepted the hidden-invariant concern: the sync-bucket fallback is now named as DEFAULT_SYNC_BUCKET instead of an inline literal.
- Accepted the route-coverage concern: added a route-level regression proving /system-bundles/channels/dev.json uses hostBundleObjectStore and does not call customerVpsObjectStore when CUSTOMER_VPS_ENABLED=true.

Commit: 21aa6d4ad
Tests:
- bun run test tests/platform/home-mirror-env-check.test.ts tests/platform/host-bundle-route.test.ts tests/platform/customer-vps-host-bundle.test.ts
- bun run typecheck
- bun run check:patterns
- git diff --check

PR #388, opened 2026-06-06
chore(platform): add managed postgres migration workflow

HamedMP:
## Summary
- add a guarded manual Platform DB Migration workflow for moving the Docker `matrixos_platform` DB into the managed Postgres URL stored in GCP Secret Manager
- keep the source dump on the VPS and ephemeral runner only; no database artifact upload
- document the migration/Cloud Run promotion gate before the Docker platform can be stopped

## Verification
- `bun run test tests/platform/platform-db-migration-workflow.test.ts`
- `python3` YAML parse for `.github/workflows/platform-db-migration.yml`
- `bun run check:patterns`
- `git diff --check`

## Invariants
- **Source of truth**: before migration, Docker Postgres `matrixos_platform` remains canonical; after a successful restore and Cloud Run verification, the GCP `platform-database-url` managed Postgres target becomes canonical for the platform control plane.
- **Lock/transaction scope**: restore mode uses `pg_restore --clean --if-exists --single-transaction` against the managed target so the target is not left partially restored on restore failure.
- **Acceptable orphan states**: verify mode only creates a VPS-local dump under `/home/deploy/backups/platform-migration/`; failed migrate runs may leave that backup file but do not promote Cloud Run or stop Docker.
- **Auth source of truth**: GitHub environment OIDC reads the GCP secret; repo `VPS_HOST`/`VPS_SSH_KEY` access the current Docker platform host as root, matching the existing Docker deploy workflow.
- **Deferred scope**: this does not migrate owner-local customer VPS Postgres databases; those remain per-owner data stores by design. It also does not automatically stop Docker after migration.

### DIFF
--- a/.github/workflows/platform-db-migration.yml
+++ b/.github/workflows/platform-db-migration.yml
@@ -81,6 +81,18 @@ jobs:
             printf 'Missing required value(s): %s\n' "${missing[*]}" >&2
             exit 1
           fi
+          safe_identifier_re='^[A-Za-z0-9_][A-Za-z0-9_-]{0,63}$'
+          safe_secret_re='^[A-Za-z0-9_][A-Za-z0-9_-]{0,127}$'
+          for name in SOURCE_CONTAINER SOURCE_DATABASE SOURCE_USER; do
+            if ! printf '%s\n' "${!name}" | grep -Eq "$safe_identifier_re"; then
+              printf 'Invalid %s: only letters, numbers, underscore, and hyphen are allowed.\n' "$name" >&2
+              exit 1
+            fi
+          done
+          if ! printf '%s\n' "$TARGET_SECRET" | grep -Eq "$safe_secret_re"; then
+            echo "Invalid TARGET_SECRET: only letters, numbers, underscore, and hyphen are allowed." >&2
+            exit 1
+          fi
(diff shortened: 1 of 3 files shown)


HamedMP:
[codex] Accepting all three Greptile findings. I will patch the migration workflow to:
- validate `source_container`, `source_database`, `source_user`, and `target_secret` with tight allowlists before any SSH/GCP command uses them;
- require `psql "$TARGET_DATABASE_URL" -c "SELECT 1"` to pass in verify mode before tolerating missing target tables on the count query;
- add a VPS-side TTL prune for old `platform-*.dump` migration snapshots so repeated verify runs cannot grow disk usage without bound.

I will push a follow-up commit and let CI/Greptile rerun.

HamedMP:
[codex] Changes since Greptile 3/5:
- Added allowlist validation for `SOURCE_CONTAINER`, `SOURCE_DATABASE`, `SOURCE_USER`, and `TARGET_SECRET` before any SSH/GCP command uses workflow inputs.
- Added an explicit managed-target `SELECT 1` connectivity check before the target table count query; verify mode no longer goes green on an unreachable DB.
- Added VPS-side TTL pruning for old `/home/deploy/backups/platform-migration/platform-*.dump` snapshots.

Commit: 49401dd95
Tests:
- bun run test tests/platform/platform-db-migration-workflow.test.ts
- python3 YAML parse for .github/workflows/platform-db-migration.yml
- bun run check:patterns
- git diff --check

PR #390, opened 2026-06-06
fix(platform): accept native app sessions

HamedMP:
Summary
- Let /api/auth/app-session accept a valid Matrix sync JWT from the native macOS shell.
- Set the existing Matrix app-session cookie from that token so app.matrix-os.com can route back into the user runtime without Clerk running inside the WebView.
- Keep Clerk token exchange as the browser path fallback.

Tests
- flox activate -- bun run test tests/platform/proxy-routing.test.ts

Invariants
- Source of truth: the Matrix sync JWT remains the runtime identity token; the platform DB is used only to verify that the JWT handle belongs to the token subject before setting the app-session cookie.
- Lock/transaction scope: no new writes or transactions; this is a request-scoped validation and cookie issuance path.
- Acceptable orphan states: if native JWT verification or runtime lookup fails, no cookie is set and the client receives a generic session-unavailable response; the Clerk browser exchange path still handles normal web auth.
- Auth source of truth: native requests authenticate via signed Matrix sync JWT; browser requests authenticate via Clerk and then mint a sync JWT as before.
- Deferred scope: no native UI changes, no provisioning changes, and no host-bundle deployment are included in this platform PR.

### DIFF
--- a/packages/platform/src/main.ts
+++ b/packages/platform/src/main.ts
@@ -3581,7 +3581,7 @@ export function createApp(deps: {
   });
 
   app.post('/api/auth/app-session', bodyLimit({ maxSize: 1024 }), async (c) => {
-    if (!platformJwtSecret || !clerkAuth) {
+    if (!platformJwtSecret) {
       applyNoStoreHeaders(c);
       return c.json({ error: 'Session unavailable' }, 503);
     }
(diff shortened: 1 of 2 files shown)


HamedMP:
* **#390** <a href="https://app.graphite.com/github/pr/HamedMP/matrix-os/390?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/HamedMP/matrix-os/390?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

PR #394, opened 2026-06-06
fix(cloud-run): bootstrap edge router secret

HamedMP:
## Summary
- Allow the Platform Cloud Run workflow to create GCP Secret Manager `edge-router-secret` from the masked GitHub environment secret `EDGE_ROUTER_SECRET` only when the GCP secret is missing.
- Scope the GitHub bootstrap secret to the single Secret Manager bootstrap step.
- Preserve fail-fast behavior when neither the GCP secret nor bootstrap secret exists, and preserve the idempotent Secret Manager accessor binding for the Cloud Run runtime service account.

## Invariants
- Source of truth: Cloud Run reads `EDGE_ROUTER_SECRET` from GCP Secret Manager; the GitHub environment secret is only a masked first-time bootstrap input when the GCP secret is absent.
- Lock/transaction scope: no database writes; the workflow either creates one Secret Manager secret and grants one IAM binding before build/deploy, or fails before deployment.
- Acceptable orphan states: if the bootstrap secret is absent and the GCP secret is absent, the workflow fails before building or deploying; if IAM cannot be granted, it fails before deployment.
- Auth source of truth: platform runtime validation still compares `X-Matrix-Edge-Secret` against the Cloud Run env var sourced from Secret Manager.
- Deferred scope: this does not perform routine secret rotation; rotation should add a new GCP Secret Manager version deliberately and update the Worker secret to match.

## Tests
- `git diff --check`

### DIFF
--- a/.github/workflows/platform-cloud-run.yml
+++ b/.github/workflows/platform-cloud-run.yml
@@ -43,7 +43,6 @@ jobs:
       CUSTOMER_VPS_IMAGE_VERSION: ${{ vars.CUSTOMER_VPS_IMAGE_VERSION }}
       POSTHOG_PUBLIC_HOST: ${{ vars.NEXT_PUBLIC_POSTHOG_HOST || 'https://eu.posthog.com' }}
       POSTHOG_PUBLIC_API_HOST: ${{ vars.NEXT_PUBLIC_POSTHOG_API_HOST || 'https://neo.matrix-os.com' }}
-      EDGE_ROUTER_SECRET_VALUE: ${{ secrets.EDGE_ROUTER_SECRET }}
     steps:
       - name: Checkout
         uses: actions/checkout@v6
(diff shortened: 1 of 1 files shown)


HamedMP:
[codex] @greptileai please review this Cloud Run edge-router secret bootstrap workflow PR at head b887ebcc745f1a41ccc5b7a36429e3e2801d5e4e.

HamedMP:
[codex] Accepting both Greptile findings. I will scope EDGE_ROUTER_SECRET to only the bootstrap step and make the step create the GCP secret only when it is missing, avoiding per-deploy version accumulation.

HamedMP:
[codex] @greptileai please review the latest head 2aec46fa311c8085cb5b9a7de954a103753e76d7 after scoping the bootstrap secret to one step and making bootstrap create-only.

PR #415, opened 2026-06-06
feat(cli): add port forwarding

Nima-Naderi:
## Summary

- Add `matrix port forward <spec>` and top-level `matrix forward <spec>` for local-to-Matrix-computer loopback forwarding.
- Add a sync-client tunnel client that binds local `127.0.0.1`, opens one authenticated WebSocket per TCP connection, bridges binary frames, caps frames/connections, and cleans up on close/error/idle timeout.
- Add gateway `/ws/forward` with authenticated bearer WebSocket access, loopback-only remote target validation, generic client-visible errors, frame/control caps, idle timeout, and shutdown cleanup.
- Document syntax, `--json` NDJSON behavior, loopback-only v1 scope, and troubleshooting in the CLI guide.

## Tests

- Passed: `pnpm --filter '@finnaai/matrix' exec tsc --noEmit`
- Passed: `pnpm --filter '@matrix-os/gateway' exec tsc --noEmit`
- Passed: `./scripts/review/check-patterns.sh` with 0 violations
- Passed: `/home/nima/matrix-os/node_modules/.bin/vitest run tests/cli/port-forward.test.ts tests/cli/unified-command-tree.test.ts tests/gateway/forward-ws.test.ts tests/gateway/auth.test.ts`
- Broad run: `/home/nima/matrix-os/node_modules/.bin/vitest run tests/cli tests/gateway` passed the new forwarding tests but failed 2 existing sandbox-dependent tests in `tests/gateway/sync/home-mirror.test.ts` because this environment denied writing `/tmp/outside-target.txt` and `/tmp/escaped-delete.txt`.

## Review/Monitoring

- Ready for CI, GitHub review, and Greptile review.
- Monitoring loop will continue until CI is green and latest trusted Greptile result is 5/5.

## Invariants

- Source of truth: forwarding state is process-local only. No persistent tunnel registry is introduced. TCP bytes remain ephemeral and are not stored.
- Lock/transaction scope: no database writes or transactions are added. Each accepted local TCP connection owns exactly one gateway WebSocket and one remote loopback TCP socket.
- Acceptable orphan states: if local TCP, WebSocket, or remote dial fails, both paired sides are closed. No durable state remains after process exit, gateway shutdown, idle timeout, or connection failure.
- Auth source of truth: gateway `authMiddleware` remains authoritative. CLI uses bearer WebSocket headers from the resolved profile token. `/ws/forward` does not accept browser query-token fallback in v1.
[...]

PR #418, opened 2026-06-06
fix(cli): harden port forwarding tunnel

Nima-Naderi:
## Summary

- Fixes the Greptile findings from #415 for CLI port forwarding tunnel lifecycle and resource handling.
- Splits local TCP reads into max-frame-sized WebSocket frames, including bytes buffered before gateway `ready`.
- Adds bounded socket-write queues/backpressure handling on CLI and gateway tunnel paths.
- Tracks pre-open gateway WebSocket connections with the same cap and idle timeout as opened tunnels.
- Fixes double-close paths in gateway failure and shutdown handling, and avoids mutating the caller's `startPortForward` options.

## Tests

- `/home/nima/matrix-os/node_modules/.bin/vitest run tests/cli/port-forward.test.ts tests/gateway/forward-ws.test.ts tests/gateway/auth.test.ts tests/cli/unified-command-tree.test.ts` (68 tests, passed; run with loopback socket permission)
- `pnpm --filter @finnaai/matrix exec tsc --noEmit`
- `pnpm --filter @matrix-os/gateway exec tsc --noEmit`
- `./scripts/review/check-patterns.sh` (0 violations; existing repo warnings only)

## Review/Monitoring

- Follow-up for Greptile comments on #415.
- Please wait for the latest Greptile review on this PR; target is 5/5 before merge.

## Invariants

- **Source of truth**: forwarding state is process-local only. The CLI listener and each gateway WebSocket connection own their sockets for the lifetime of the tunnel; there is no persistence or registry beyond bounded in-memory connection sets.
- **Lock/transaction scope**: no database writes or transactions are involved. Socket and WebSocket cleanup is guarded by per-connection closed flags or active-set deletion.
- **Acceptable orphan states**: if either TCP or WebSocket side fails, the paired side is closed and pending buffers/timers are cleared. Pre-open gateway WebSockets are capped, timed out, and cleaned up before promotion to active tunnel state.
- **Auth source of truth**: `/ws/forward` remains authenticated by the existing gateway bearer-token WebSocket path; this PR does not add query-token auth.
- **Deferred scope**: still forward-only. No reverse tunnels, multiplexing, public bind address, public preview URL, persistent tunnel registry, or shareable HTTPS publishing.

### DIFF
--- a/packages/gateway/src/forward-ws.ts
+++ b/packages/gateway/src/forward-ws.ts
@@ -5,6 +5,10 @@ export const FORWARD_WS_MAX_CONTROL_BYTES = 2048;
 export const FORWARD_WS_MAX_FRAME_BYTES = 64 * 1024;
 export const FORWARD_WS_IDLE_TIMEOUT_MS = 60_000;
 export const FORWARD_WS_MAX_CONNECTIONS = 100;
+const FORWARD_WS_MAX_PENDING_SOCKET_WRITE_BYTES = 256 * 1024;
+const FORWARD_WS_BUFFER_HIGH_WATER_BYTES = 256 * 1024;
+const FORWARD_WS_BUFFER_LOW_WATER_BYTES = 128 * 1024;
+const FORWARD_WS_BACKPRESSURE_POLL_MS = 10;
 
 const ForwardPortSchema = z.number().int().min(1).max(65_535);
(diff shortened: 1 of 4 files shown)


Nima-Naderi:
@greptileai please review this follow-up for #415 and verify the fixes are 5/5 confident.

PR #438, opened 2026-06-07
fix(086): exchange native token before loading shell

HamedMP:
## Summary
- exchange the macOS principal token with /api/auth/app-session using URLSession before loading the hosted shell in WKWebView
- require the Matrix app-session cookie before loading the shell, otherwise return to native sign-in instead of showing Clerk web login
- add AppTests for request construction and cookie validation

## Invariants
- Source of truth: the native principal token remains the macOS app auth source of truth; /api/auth/app-session converts it into a scoped app-domain cookie for hosted shell rendering only.
- Lock/transaction scope: no database writes or locks in this native client change; the exchange is a single bounded HTTP request followed by cookie installation in the WebView store.
- Acceptable orphan states: if cookie installation or exchange fails, the app shows the native sign-in state and does not load the Clerk web auth page in the embedded WebView.
- Auth source of truth: platform-issued native principal JWT first; Clerk web session remains platform-side and is not initiated inside WKWebView.
- Deferred scope: this does not migrate the macOS app to ClerkKit/native OAuth. That should be a separate PR if we decide to replace the device-code bridge with Clerk's native SDK flow.

## Verification
- swift test --package-path macos --filter AppTests
- swift test --package-path macos

### DIFF
--- a/macos/Sources/App/WebShellView.swift
+++ b/macos/Sources/App/WebShellView.swift
@@ -143,43 +143,20 @@ private struct WebShellView: NSViewRepresentable {
         coordinator.lastBearerToken = bearerToken
         coordinator.lastRequestedURL = url
         coordinator.destinationURL = url
-        guard let bearerToken, !bearerToken.isEmpty, let request = appSessionExchangeRequest(for: url, token: bearerToken) else {
-            coordinator.exchangeInFlight = false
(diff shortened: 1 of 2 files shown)


HamedMP:
@greptileai

PR #441, opened 2026-06-07
fix(086): stop native login refresh loop

HamedMP:
## Summary
- Prevent the macOS native shell from looping when the hosted shell redirects to sign-in.
- Keep the native sign-in prompt visible until an explicit native sign-in completes.
- Add regression coverage for the native web shell auth state transition.

## Tests
- `swift test --package-path macos --filter AppModelAuthTests`
- `swift test --package-path macos`
- `./script/build_and_run.sh --verify`

## Invariants
- **Source of truth**: the native principal token remains the Keychain-backed Matrix token; the hosted app-session cookie is derived from that token by `/api/auth/app-session`.
- **Lock/transaction scope**: no database writes or transactions are introduced; this is local SwiftUI/WebView auth state only.
- **Acceptable orphan states**: if the hosted shell rejects the derived session, the native app stays on the sign-in prompt while preserving the existing token for debugging/retry; it does not reload the hosted login page in a loop.
- **Auth source of truth**: explicit native device sign-in is the only event that clears a hosted-auth rejection. Automatic keychain reloads cannot mark a previously rejected token valid again.
- **Deferred scope**: this does not change Clerk, platform routing, device auth issuance, or customer VPS gateway behavior.

### DIFF
--- a/macos/Sources/App/WebShellView.swift
+++ b/macos/Sources/App/WebShellView.swift
@@ -9,9 +9,7 @@ struct MatrixWebShellPanel: View {
     let url: URL?
     let title: String
 
-    @State private var token: String?
-    @State private var didResolveToken = false
-    @State private var authRequired = false
+    @State private var authState = WebShellAuthState()
 
     var body: some View {
         VStack(spacing: 0) {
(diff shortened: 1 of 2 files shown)


HamedMP:
Closing as superseded by #443 and the follow-up native workspace work. The native login refresh-loop fix is covered by the newer WebShellAuthState/session exchange work, and this branch is now stale/conflicted.

PR #442, opened 2026-06-07
fix(symphony): compile status dashboard truncation

HamedMP:
## Summary
- Move `String.length/1` out of the `truncate/2` guard in the Symphony status dashboard.
- Add a regression assertion so the release compiler failure cannot be reintroduced.

## Invariants
- Source of truth: terminal dashboard formatting remains local to `StatusDashboard`; no runtime state source changes.
- Lock/transaction scope: no database or file writes are introduced.
- Acceptable orphan states: unchanged; this only affects dashboard string truncation during Symphony startup/build.
- Auth source of truth: unchanged; no auth path or external API path changes.
- Deferred scope: this PR does not change Matrix app service controls or deploy hamedmp directly; it unblocks the host-bundle release failure from run 27103622889.

## Verification
- `bun run test tests/deploy/customer-vps/symphony-systemd.test.ts`
- `bun run typecheck`
- `bun run check:patterns`

### DIFF
--- a/packages/symphony-elixir/lib/symphony_elixir/status_dashboard.ex
+++ b/packages/symphony-elixir/lib/symphony_elixir/status_dashboard.ex
@@ -1934,8 +1934,12 @@ defmodule SymphonyElixir.StatusDashboard do
   defp alternate_key(key) when is_atom(key), do: Atom.to_string(key)
   defp alternate_key(key), do: key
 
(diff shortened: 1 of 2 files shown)


HamedMP:
Closing as superseded by main commit 46c36308, which contains the same Symphony status-dashboard guard compile fix and has a successful host-bundle release.

PR #445, opened 2026-06-08
fix(symphony): export host env to runtime service

HamedMP:
## Summary
- add a dedicated least-privilege /opt/matrix/env/symphony.env containing only MATRIX_HANDLE, PLATFORM_INTERNAL_URL, and UPGRADE_TOKEN for the host-managed Symphony runtime
- pin matrix-gateway.service to SYMPHONY_PORT=4766 so stale legacy runner config cannot misroute the proxy
- generate symphony.env during customer VPS bootstrap and existing VPS host-bundle updates before enabling bundled units
- add static regression coverage for the shipped unit, cloud-init copy, and sync-agent update path

## Invariants
- Source of truth: platform host env remains the source of truth for machine metadata and tokens; symphony.env is a derived least-privilege runtime env for Symphony only.
- Lock/transaction scope: no database writes or runtime transactions are changed; this is systemd environment wiring and update-time env-file generation only.
- Acceptable orphan states: an existing host may still need a unit reinstall or service restart before the new environment applies; once the host bundle installs units, systemd owns restart behavior.
- Auth source of truth: Matrix gateway auth and platform bridge token validation are unchanged; Symphony receives only the bridge values it already needs.
- Deferred scope: this does not change Symphony issue selection, Linear credentials, or Codex agent execution behavior.

## Verification
- bun run test tests/deploy/customer-vps/symphony-systemd.test.ts
- bun run test tests/platform/customer-vps-cloud-init.test.ts
- bun run check:patterns

### DIFF
--- a/distro/customer-vps/cloud-init.yaml
+++ b/distro/customer-vps/cloud-init.yaml
@@ -79,6 +79,7 @@ write_files:
       User=matrix
       Group=matrix
       EnvironmentFile=/opt/matrix/env/host.env
+      Environment=SYMPHONY_PORT=4766
       WorkingDirectory=/opt/matrix
       ExecStart=/opt/matrix/bin/matrix-gateway
       Restart=on-failure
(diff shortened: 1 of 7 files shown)


HamedMP:
@greptileai please re-review the latest head 8c717e713bc8a37b229ab76ae6abb251165e127e after the least-privilege symphony.env fix.

PR #454, opened 2026-06-09
fix(macos): preserve native shell session cookie

HamedMP:
fix(macos): preserve native shell session cookie

fix(macos): stop hosted-shell auth loop without signing out native session

When the hosted web shell at app.matrix-os.com redirected an authenticated
native request to /login, the app retried the app-session exchange once and
then signed out the native principal (clearSessionAfterHostedAuthFailure ->
signOutNow). That wiped a valid device-auth token, dropped the user to
onboarding, and looped: re-sign-in -> workspace -> Home -> /login -> sign-out.

The native principal and gateway session are valid; only the hosted *web*
session needs re-auth. Replace the destructive path with a non-destructive
markHostedShellAuthRequired() that flags the hosted shell needs sign-in while
keeping the native session. WebShellAuthState already caps automatic retries at
one, so the Home panel settles on the native sign-in prompt with no loop.

Also keeps the verified ASWebAuthenticationSession completion-handler fix
(nonisolated factory + MainActor hop) that resolves the post-approval crash.

Tests: 192 macOS tests pass. Telemetry confirms one retry then a stable
"keeping native session (no sign-out)" with a single app instance.

fix(macos): persist hosted-shell data store, strip stale Clerk cookies

Revert the non-persistent WKWebView data store (it forced a full Home-tab
reload every visit). The real /login root cause was the shell proxy path, now
fixed. Keep clearInterferingClerkCookies as a per-load guard so leftover Clerk
client cookies cannot override the native app session.

feat(macos): real code editor with content, highlighting, and find

Fix the empty-editor bug: applyLayoutPreferences ran before the view had
geometry, so the text container width was zero and content never laid out
(only the gutter rendered). Defer the first content apply until the scroll
view has real width. Add per-language syntax highlighting (Swift, TS/JS, Go,
Rust, etc.), number/block-comment coloring, right-aligned line numbers, and a
native in-file find bar (Cmd+F/Cmd+G, case toggle, match count).

fix(macos): clean terminal rendering

Drop the zero-frame initial resize that connected the PTY at a default size
then reflowed; set caret and selection colors for the dark surface; replace the
never-installed Nerd Font fallback list with IBM Plex Mono/Menlo at 13pt; clamp
reported cols/rows to [1,500] before forwarding to the session.

fix(macos): editor fills its pane instead of a thin centered column

EditorScrollHost only laid out on the first pass; SwiftUI resizes the host by
setting its frame, which does not trigger layout() on a non-autoresizing view,
so the scroll view stayed latched at its initial narrow width. Re-lay contents
on setFrameSize, and stop capping the text view's max width at that early
contentSize (which pinned dense files like JSON to one character per line).

fix(macos): debounce terminal resize to settle full-size grid

SwiftTerm emits many intermediate grid sizes during initial layout and window
drags; forwarding each made zellij thrash and could leave it stuck at an early,
smaller-than-the-view grid. Coalesce resizes with a 90ms settle so only the
final full size reaches the server (re-sent after attach by the WS client).

feat(macos): keep terminals alive across switches + Meslo glyphs

Mount every live (cached) terminal session at once and show only the active one
(opacity + hit-testing), so switching sessions is instant: inactive terminals stay
alive with their PTY stream and scrollback instead of tearing down and re-attaching
from scratch. Gate keyboard focus to the foreground terminal, and force-focus the
newly-active one on switch so input lands immediately. Restore Nerd Font priority
(MesloLGS NF first) so zellij powerline/glyph icons render, with Plex Mono/Menlo
fallback. Subtle crossfade on switch.

feat(macos): keep hosted web shell mounted like a browser tab

Home was a switch arm, so leaving and returning tore down the WKWebView and
forced a full reload + re-auth exchange. Keep MatrixWebShellPanel mounted at all
times (opacity + hit-testing gated on the active section) so the shell persists
its page, scroll, and auth state across section switches; other sections render
on top when active.

fix(macos): restore editor content (revert max-width uncap)

The earlier max-width uncap (maxSize=greatest / minSize=0) regressed dense files
to a blank pane (only the gutter rendered). The scroll-view relayout on
setFrameSize is what actually fixes the narrow-column bug, so keep that and
restore the bounded maxSize/minSize that lays content out correctly.

fix(macos): revert editor setFrameSize override that blanked all files

The setFrameSize override (1fb41befd) latched the scroll view at a bad frame on
SwiftUI-driven resizes, leaving every file rendered as an empty pane (gutter only).
Restore the original layout() so content renders again. The full-width layout
polish will be redone correctly.

fix(macos): revert keep-alive that blanked editor and piled up zellij clients

The keep-alive experiment caused two regressions: the always-mounted hosted-shell
WKWebView overlapped the editor's NSView and blanked it, and mounting every cached
terminal session opened a WebSocket client per session (plus leaks on remount),
piling clients onto one zellij session and making it laggy. Restore the single
active terminal + plain section switch (snappy, one client, like the web shell).
Keep the resize debounce and restore the Meslo Nerd Font for zellij glyphs.

feat(macos): upgrade native workspace surfaces

Summary:
- Replace the native editor wrapper with CodeEditSourceEditor and copy SwiftPM resource bundles into the app bundle so language assets are present at runtime.
- Add Cmd+O file search, an AppKit file navigator, multi-pane task layout, pane toggles, and Cmd+1 through Cmd+9 global tab focus.
- Stabilize terminal startup with attach/resize settling, shared named session clients, safer terminal error decoding, and project-scoped task terminal cwd.

Rationale:
- The previous native shell relied on hand-rolled editor, tree, and pane controls that made common task workflows feel incomplete.
- Terminal startup needed to avoid resize bursts and duplicate named-session sockets while still preserving tab reuse.

Tests:
- swift test --package-path macos --filter AppModelAuthTests --filter ShellMessageCodecTests --filter TerminalRendererTests
- swift test --package-path macos
- ./script/build_and_run.sh run

### DIFF
--- a/macos/Package.resolved
+++ b/macos/Package.resolved
@@ -1,6 +1,42 @@
 {
-  "originHash" : "76df76a2d249c98749db73d482ec32c708d4b71e92c369aa06edf4ebff59ff47",
+  "originHash" : "103d0ccde6a5a4ef99bf2a2965a555af2a03fdf2f1779b53a7453cf4a6d45af5",
   "pins" : [
+    {
+      "identity" : "codeeditlanguages",
+      "kind" : "remoteSourceControl",
+      "location" : "https://github.com/CodeEditApp/CodeEditLanguages.git",
+      "state" : {
+        "revision" : "331d5dbc5fc8513be5848fce8a2a312908f36a11",
+        "version" : "0.1.20"
+      }
+    },
+    {
+      "identity" : "codeeditsourceeditor",
+      "kind" : "remoteSourceControl",
+      "location" : "https://github.com/CodeEditApp/CodeEditSourceEditor",
+      "state" : {
+        "revision" : "412b0a26cbeb3f3148a1933dd598c976defe92a6",
+        "version" : "0.12.0"
+      }
+    },
+    {
+      "identity" : "codeedittextview",
+      "kind" : "remoteSourceControl",
+      "location" : "https://github.com/CodeEditApp/CodeEditTextView.git",
+      "state" : {
+        "revision" : "a5912e60f6bac25cd1cdf8bb532e1125b21cf7f7",
+        "version" : "0.10.1"
+      }
+    },
+    {
+      "identity" : "rearrange",
+      "kind" : "remoteSourceControl",
+      "location" : "https://github.com/ChimeHQ/Rearrange",
+      "state" : {
+        "revision" : "4de8be41dba304192e87dc0a11e0aa39e72aa2e8",
+        "version" : "2.1.1"
+      }
+    },
     {
       "identity" : "swift-argument-parser",
       "kind" : "remoteSourceControl",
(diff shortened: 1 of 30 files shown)


HamedMP:
* **#454** <a href="https://app.graphite.com/github/pr/HamedMP/matrix-os/454?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/HamedMP/matrix-os/454?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

HamedMP:

Status: draft and conflicts with `main`; this branch contains a much broader native macOS stack than the session-cookie fix in its title. Native shell, terminal, workspace, and auth code have all moved.

To revive: extract a minimal failing cookie/session test against the current native WebShell/auth callback path, implement only that fix in a fresh branch, and leave unrelated macOS changes out. Owner keep/kill decision requested.

PR #465, opened 2026-06-10
fix(platform): send CLI device billing failures to setup

Nima-Naderi:
## Summary
- Send CLI device-auth `402 billing_required` responses into the setup/billing handoff instead of the generic refresh state.
- Preserve validated `/auth/device` return paths through the platform auth page, Stripe checkout, and final app-session redirect.
- Keep auth-page checkout return overrides scoped to device-auth flows so configured Stripe success/cancel URLs still apply to normal auth-page checkout.

## Tests
- `bun run test tests/platform/device-routes.test.ts`
- `bun run test tests/platform/proxy-routing.test.ts -t "only preserves the runtime selector after auth|does not persist a selected runtime slot|keeps legacy no-container app-domain users"`
- `bun run typecheck`
- `bun run check:patterns`
- `git diff --check`

## Review/Monitoring
- Ready-for-CI label applied.
- Addressed Greptile's `4/5` finding about unconditional checkout return-path overrides by adding `returnPath` only when a validated device return target exists.

## Invariants
- Source of truth: platform Postgres remains the source for billing entitlement, runtime identity, and active machine state; this change only adjusts browser handoff behavior around existing `/api/auth/app-session` and `/billing/checkout` decisions.
- Lock/transaction scope: no database writes or transaction boundaries changed; the route still delegates provisioning/session decisions to existing platform services.
- Acceptable orphan states: if checkout/provisioning fails or billing has not propagated, the existing auth page retry states remain in place; no device code is approved until `/auth/device/approve` succeeds.
- Auth source of truth: Clerk session tokens remain the browser auth source; device approval still requires Clerk bearer auth plus CSRF, and device return paths are restricted to same-origin `/auth/device`.
- Deferred scope: this PR does not change billing plan selection, Stripe webhook timing, or VPS provisioning semantics.

### DIFF
--- a/packages/platform/src/auth-routes.ts
+++ b/packages/platform/src/auth-routes.ts
@@ -375,8 +375,8 @@ function approvalPage(
           showConfirm();
           return;
         }
-        if (res.status === 404) {
-          if (nativeApp) {
+        if (res.status === 402 || res.status === 404) {
+          if (nativeApp && res.status === 404) {
             showRuntimeSetupState();
             return;
           }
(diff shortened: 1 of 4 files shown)


HamedMP:
Closing/reopening so the PR object is authored by @HamedMP now that GitHub auth has switched. Replacement PR uses the same branch and Hamed-authored head commit.

PR #478, opened 2026-06-10
fix(shell): allow billing setup shell entry

Nima-Naderi:
## Summary
- Allow `/?billing=setup` through the shell proxy as a public pre-VPS entry point so the BillingGate can render instead of Clerk redirecting to sign-in.
- Keep the normal `/` shell route protected.
- Add route-classification and proxy-level regression coverage for the exact candidate smoke failure.

## Evidence
- Merged PR #477 fixed the anonymous `/sign-in` auth-shell path, but the main Platform Cloud Run candidate `matrix-platform-00075-ruq` failed smoke with: `Candidate pre-VPS billing shell did not serve the settings billing gate.`
- Direct candidate probe showed `/?billing=setup` returned `307 Location: http://app.matrix-os.com/sign-in?...`, so Next/Clerk redirected before the billing gate rendered.

## Invariants
- Source of truth: shell proxy route classification controls which pre-VPS shell entry points may render without Clerk middleware redirecting first.
- Lock/transaction scope: no persistence or transactions changed.
- Acceptable orphan states: none introduced; a missing/invalid Clerk session still cannot access the normal shell, API, files, apps, modules, or websockets.
- Auth source of truth: Clerk remains the auth source for protected shell routes; only the billing setup entry HTML is public so its client-side billing/auth gate can render.
- Deferred scope: this does not change Stripe entitlement decisions, platform `/api/auth/app-session` behavior, or customer VPS routing.

## Validation
- `/home/deploy/matrix-os/node_modules/.bin/vitest run tests/shell/proxy-auth.test.ts`
- `pnpm --filter ./shell exec tsc --noEmit`
- `bun run check:patterns`
- `/home/deploy/matrix-os/node_modules/.bin/vitest run tests/shell/proxy-auth.test.ts tests/platform/ci-workflows.test.ts`

Note: this worktree could not run a fresh `pnpm install` because the host is inode-exhausted (`df -i` showed 100% inode use); targeted tests used the root checkout's existing installed toolchain via a temporary symlink that was removed before commit.

### DIFF
--- a/shell/src/lib/proxy-routes.ts
+++ b/shell/src/lib/proxy-routes.ts
@@ -1,9 +1,11 @@
-export function isPublicShellPath(pathname: string): boolean {
+export function isPublicShellPath(pathname: string, search = ""): boolean {
+  const searchParams = new URLSearchParams(search.startsWith("?") ? search.slice(1) : search);
   return (
     pathname === "/health" ||
     pathname === "/manifest.json" ||
     pathname === "/og.png" ||
     pathname === "/favicon.ico" ||
+    (pathname === "/" && searchParams.get("billing") === "setup") ||
     pathname === "/sign-in" ||
     pathname.startsWith("/sign-in/") ||
     pathname === "/sign-up" ||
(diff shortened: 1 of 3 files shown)


Nima-Naderi:
Closing to recreate under the HamedMP GitHub account; branch and commit are unchanged.

PR #487, opened 2026-06-11
fix(platform): self-heal stuck provisioning and unblock failed-machine retries

HamedMP:
## Summary

Spec 092 **Phase A** (provisioning reliability — first PR in the onboarding-unification Graphite stack). Fixes the two failure modes that strand paying users at the moment of highest intent, with no UI change:

1. **Stuck-forever provisioning.** `reconcileProvisioning()` previously refreshed IP metadata for a booted-but-never-registered machine and left it `provisioning` indefinitely. It now transitions such a machine to `failed` with `registration_timeout` once its registration token expires, and reaps the abandoned Hetzner server via the existing deletion queue.
2. **Failed rows permanently blocking re-provisioning.** A `failed` row (with `deleted_at IS NULL`) held the active `(clerk_user_id, runtime_slot)` unique slot, so `provision()` hit `activeProvisionResponse` → 409 forever. `provision()` now retires the failed row (soft-delete) and inserts a fresh attempt **in the same transaction**, so the unique slot is satisfied at every instant and the user can self-retry. Retries are bounded by `CUSTOMER_VPS_MAX_PROVISION_ATTEMPTS` (default 3); past the cap, provision returns `retry_exhausted` (409) for a support escalation.

Adds `user_machines.attempt` (int, default 1) and a `retireUserMachine()` DB helper. `failure_code`/`failure_at` already existed.

Deferred to later phases (declared so there is no dead code): the `/api/journey` endpoint, `billing_checkout_attempts`/settling, and `recover()` work (verified already atomic — single row mutated in place inside `runInPlatformTransaction`, so no change needed).

## Tests

New `tests/platform/customer-vps-reliability.test.ts` (7 cases, TDD red→green): registration-timeout reconcile + reap, still-valid-token not failed, failed-row retire+retry, concurrent-retry convergence to one live row, attempt cap, retired-row registration rejection, expired-token registration rejection.

- `vitest run` new file: 7/7 pass
- `vitest run` existing `customer-vps*` + `db` suites: 72/72 pass (no regression)
- `tsc --noEmit` (platform `tsconfig.typecheck.json`): clean
- `scripts/review/check-patterns.sh`: 0 violations (5 pre-existing gateway warnings, not in this diff)

## Invariants

- **Source of truth**: `user_machines` is canonical for machine lifecycle. The active machine for a `(clerk_user_id, runtime_slot)` is the single row with `deleted_at IS NULL` (enforced by `idx_user_machines_clerk_slot_active`). A `failed` row is non-routable but still occupies the slot until retired.
- **Lock/transaction scope**: retire-failed + insert-new-attempt runs inside `runInPlatformTransaction` under the per-user `pg_advisory_xact_lock` (`lockUserMachineProvisioning`) when billing is enabled. The Hetzner `createServer` call happens *after* the row insert (existing flow); the retired server's deletion is queued *outside* the transaction. No network call occurs inside the transaction.
- **Acceptable orphan states**: a retired failed machine's Hetzner server may briefly outlive its soft-deleted DB row — it is queued for reaping (`failed_retry_retire`) and drained by the reconciler's existing `retryProviderDeletions()`. Double-queueing a server already reaped during `registration_timeout` is harmless (idempotent delete).
- **Auth source of truth**: unchanged. `/vps/register` still requires a valid, unexpired registration token matching the row; retired (soft-deleted, `status='failed'`) rows are rejected by the existing status + `deleted_at IS NULL` guards.
- **Deferred scope**: journey endpoint, checkout-attempt/settling records, and shell/CLI/mobile rendering ship in Phases B–E of the stack.

### DIFF
--- a/packages/platform/src/customer-vps-config.ts
+++ b/packages/platform/src/customer-vps-config.ts
@@ -23,6 +23,7 @@ export interface CustomerVpsConfig {
   registrationTokenTtlMs: number;
   reconciliationBatchSize: number;
   reconciliationStaleAfterMs: number;
+  maxProvisionAttempts: number;
 }
 
 function numberFromEnv(value: string | undefined, fallback: number): number {
@@ -67,5 +68,6 @@ export function loadCustomerVpsConfig(env: NodeJS.ProcessEnv = process.env): Cus
     registrationTokenTtlMs: numberFromEnv(env.CUSTOMER_VPS_REGISTRATION_TOKEN_TTL_MS, 15 * 60 * 1000),
     reconciliationBatchSize: numberFromEnv(env.CUSTOMER_VPS_RECONCILIATION_BATCH_SIZE, 50),
     reconciliationStaleAfterMs: numberFromEnv(env.CUSTOMER_VPS_RECONCILIATION_STALE_AFTER_MS, 10 * 60 * 1000),
+    maxProvisionAttempts: numberFromEnv(env.CUSTOMER_VPS_MAX_PROVISION_ATTEMPTS, 3),
   };
 }
--- a/packages/platform/src/customer-vps-errors.ts
+++ b/packages/platform/src/customer-vps-errors.ts
@@ -7,6 +7,8 @@ export type CustomerVpsFailureCode =
   | 'billing_required'
   | 'not_found'
   | 'registration_rejected'
+  | 'registration_timeout'
+  | 'retry_exhausted'
   | 'unknown';
 
 export class CustomerVpsError extends Error {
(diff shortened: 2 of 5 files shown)


HamedMP:
* **#500** <a href="https://app.graphite.com/github/pr/HamedMP/matrix-os/500?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#495** <a href="https://app.graphite.com/github/pr/HamedMP/matrix-os/495?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#494** <a href="https://app.graphite.com/github/pr/HamedMP/matrix-os/494?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#487** <a href="https://app.graphite.com/github/pr/HamedMP/matrix-os/487?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/HamedMP/matrix-os/487?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

PR #549, opened 2026-06-15
fix(shell): keep Clerk sign-up on app.matrix-os.com, not the Account Portal

HamedMP:
## Summary

On `app.matrix-os.com/sign-in`, the "Sign up" link redirected to Clerk's hosted
Account Portal (`accounts.matrix-os.com`) instead of `app.matrix-os.com/sign-up`.

Root cause: `<ClerkProvider>` had no `signInUrl`/`signUpUrl`. The shell reads
`NEXT_PUBLIC_CLERK_SIGN_IN_URL`/`NEXT_PUBLIC_CLERK_SIGN_UP_URL`, but those are
never baked into the build — the deployment sets the **un-prefixed**
`CLERK_SIGN_IN_URL`/`CLERK_SIGN_UP_URL` (in `distro/docker-compose.platform.yml`),
which neither Clerk's SDK nor the shell code reads. With no app URLs, Clerk's
components fall back to the Account Portal.

Fix: set `signInUrl`/`signUpUrl` on `<ClerkProvider>` (honoring the env if ever
set, defaulting to `/sign-in` and `/sign-up`).

## Scope / deploy

- Shell-only, 1 file. Reaches `app.matrix-os.com` after a host-bundle rebuild +
  deploy (`NEXT_PUBLIC_*` is build-time).
- Immediate (no-deploy) relief is also available via the Clerk dashboard Account
  Portal paths; this PR is the durable in-repo fix.
- Separate follow-up: pass `NEXT_PUBLIC_CLERK_SIGN_*` as build args
  (cloudbuild/host-bundle) and drop the dead un-prefixed `CLERK_SIGN_*` env.

## Tests

react-doctor clean on the change (only a pre-existing `@clerk/nextjs`
supply-chain note); `check:patterns` clean. Symptom #1 (Cloudflare/Turnstile
challenge not loading) was a Clerk **bot-protection** dashboard setting, resolved
separately.

### DIFF
--- a/shell/src/app/layout.tsx
+++ b/shell/src/app/layout.tsx
@@ -61,8 +61,12 @@ export default async function RootLayout({
 
   return (
     <ClerkProvider
-      signInUrl={process.env.NEXT_PUBLIC_CLERK_SIGN_IN_URL ?? "/sign-in"}
-      signUpUrl={process.env.NEXT_PUBLIC_CLERK_SIGN_UP_URL ?? "/sign-up"}
+      // Ground Clerk's sign-in/sign-up component cross-links to the app's own
+      // routes; without these, Clerk falls back to its hosted Account Portal
+      // (accounts.matrix-os.com). These paths are fixed app routes
+      // (shell/src/app/sign-in, sign-up).
+      signInUrl="/sign-in"
+      signUpUrl="/sign-up"
     >
       <html
         lang="en"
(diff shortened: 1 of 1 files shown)


HamedMP:
Duplicate of #550, which has the same ClerkProvider signInUrl/signUpUrl fix plus a regression test. Closing in favor of #550.

PR #552, opened 2026-06-16
fix(vps): preflight disk space before host updates

HamedMP:
## Summary
- add a sync-agent disk headroom preflight before downloading/extracting host bundle updates
- derive required free space from manifest bundle size: size times 8 plus 1GiB buffer, and check the staging filesystem where download/extract writes
- clean stale legacy /opt/matrix-upgrade bundle artifacts before checking headroom
- cover the updater behavior in the host-bundle regression test

## Invariants
- Source of truth: /opt/matrix/app/BUNDLE_VERSION and /opt/matrix/release.json remain the installed runtime source of truth; release metadata still comes from platform bundle manifests.
- Lock/transaction scope: no database writes; the critical install section still starts only after bundle download, checksum, extraction, and validation succeed. The new disk preflight runs before download and before services are stopped.
- Acceptable orphan states: if headroom is insufficient, the update trigger fails safely and leaves the current app/services intact for a later retry after cleanup.
- Auth source of truth: unchanged; sync-agent still uses host.env PLATFORM_INTERNAL_URL/UPGRADE_TOKEN and local service auth remains unchanged.
- Deferred scope: this does not shrink host bundle size or implement fleet-wide disk cleanup; it prevents low-disk upgrades from destabilizing services and removes stale legacy upgrade tarballs.

## Validation
- bash -n distro/customer-vps/host-bin/matrix-sync-agent
- bun run check:patterns
- git diff --check

### DIFF
--- a/distro/customer-vps/host-bin/matrix-sync-agent
+++ b/distro/customer-vps/host-bin/matrix-sync-agent
@@ -24,6 +24,10 @@ readonly HEALTH_URL="http://127.0.0.1:4000/health"
 readonly AUTO_APPLY_FAILED_MARKER="$APP_DIR/.auto-apply-failed"
 readonly STAGING_ARTIFACT_TTL_SECONDS="${MATRIX_STAGING_ARTIFACT_TTL_SECONDS:-86400}"
 readonly STAGING_CLEANUP_INTERVAL_SECONDS="${MATRIX_STAGING_CLEANUP_INTERVAL_SECONDS:-3600}"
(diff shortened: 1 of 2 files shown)


HamedMP:

Closing as superseded. Current `main` performs update headroom checks in `matrix-sync-agent`, emits a bounded `insufficient_disk_space` state, and cleans repairable staging artifacts. The replacement landed in [246efb9](https://github.com/HamedMP/matrix-os/commit/246efb90ac61a1944cc8f0cfe6663ff6bc99da00) and was subsequently expanded in the current updater.

PR #561, opened 2026-06-16
fix(shell): use owner handle in terminal prompt

HamedMP:
Summary
- Generate a Matrix terminal prompt-label helper beside the Zellij config.
- Read the owner-controlled ~/system/handle.json at shell startup and prefer a safe handle over displayName.
- Keep MATRIX_TERMINAL_PROMPT as an explicit override and fall back to the OS user when no safe owner label exists.

Validation
- pnpm exec vitest run tests/gateway/shell-zellij.test.ts tests/gateway/zellij-runtime.test.ts
- pnpm exec tsc --noEmit -p packages/gateway/tsconfig.json
- bun run check:patterns
- git diff --check

Invariants
- Source of truth: owner-controlled ~/system/handle.json remains canonical for the user-facing prompt label; the Linux account and VPS hostname remain internal runtime identity.
- Lock/transaction scope: no database writes or multi-step persistent mutations beyond the existing atomic config-file writes.
- Acceptable orphan states: if handle.json is missing, malformed, or unsafe, the shell falls back to the OS user prompt; existing terminal sessions keep their current prompt until restarted.
- Auth source of truth: no auth change; this only reads local owner files inside the user home.
- Deferred scope: no settings UI for editing the handle/display name in this PR; this only makes the terminal consume the existing owner identity file safely.

### DIFF
--- a/packages/gateway/src/shell/zellij-config.ts
+++ b/packages/gateway/src/shell/zellij-config.ts
@@ -5,6 +5,7 @@ export type MatrixZellijConfigPaths = {
   file: string;
   shellFile: string;
   bashrcFile: string;
+  promptLabelFile: string;
   layoutDir: string;
   layoutFile: string;
 };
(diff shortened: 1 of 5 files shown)


HamedMP:
[codex] Accept. I’ll make ENOENT the only quiet expected case and log unexpected identity-read failures to stderr; normal shell launch already redirects that stderr, while direct debugging gets the signal.

PR #570, opened 2026-06-17
fix(runtime): parallelize incremental bundle object uploads

HamedMP:
## Summary
- Parallelize content-addressed incremental object uploads in the host-bundle publisher.
- Treat R2 precondition/existing-object responses as success for SHA-addressed objects.
- Increase the publish job timeout from 10 to 20 minutes after the first incremental publish hit the timeout.

## Invariants
- Source of truth: R2 immutable bundle/object keys plus platform release metadata remain canonical.
- Lock/transaction scope: publish still writes immutable objects first, then registers the release; no DB transaction wraps object-store uploads.
- Acceptable orphan states: uploaded bundle/object bytes may exist without release registration and are safe because keys are immutable/content-addressed.
- Auth source of truth: platform registration still uses PLATFORM_SECRET via header file; R2 auth stays in GitHub secrets.
- Deferred scope: VPS-side delta apply remains disabled by requiresFullBundle until the installer/apply layer lands.

## Verification
- bash -n scripts/publish-release.sh
- git diff --check
- bun run check:patterns

### DIFF
--- a/.github/workflows/host-bundle-release.yml
+++ b/.github/workflows/host-bundle-release.yml
@@ -277,7 +277,7 @@ jobs:
     needs: [dev-bundle-gate, build]
     if: needs.dev-bundle-gate.outputs.should_build == 'true'
     runs-on: ubuntu-latest
-    timeout-minutes: 10
+    timeout-minutes: 20
     env:
       R2_ACCOUNT_ID: ${{ secrets.R2_BUNDLES_ACCOUNT_ID || secrets.R2_ACCOUNT_ID }}
       AWS_ACCESS_KEY_ID: ${{ secrets.R2_BUNDLES_ACCESS_KEY_ID || secrets.R2_ACCESS_KEY_ID }}
--- a/scripts/publish-release.sh
+++ b/scripts/publish-release.sh
@@ -157,6 +157,27 @@ for entry in manifest.get('files', []):
 " "$INCREMENTAL_MANIFEST"
 }
 
+validate_incremental_object_list() {
+  local object_sha256 object_size object_key object_file actual_size actual_sha256
+  while IFS=$'\t' read -r object_sha256 object_size object_key; do
+    object_file="$DIST_DIR/objects/sha256/$object_sha256"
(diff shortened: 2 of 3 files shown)


HamedMP:
[codex] Accepting Greptile feedback. I’m going to make the wait failure propagation explicit, split local object validation from background upload so validation failures cannot leave orphan upload jobs, and broaden 412 detection for R2/AWS CLI precondition responses. I’ll push a follow-up commit and let Greptile/CI rerun on the new head.

HamedMP:
[codex] Changes since Greptile review:
- Made slot waiting explicit with wait -n || return.
- Split local incremental object validation into validate_incremental_object_list before any background upload starts, so validation failures cannot orphan upload jobs.
- Broadened 412/precondition detection with a numeric 412 fallback while still logging raw stderr for non-matches.
Commits: 64c63dd18d
Tests: bun run test tests/platform/customer-vps-host-bundle.test.ts; bash -n scripts/publish-release.sh; git diff --check.

HamedMP:
[codex] Accepting the latest Greptile finding. The prior implementation used shell job introspection, which can miss jobs because the count runs in command substitution; I’m replacing it with an explicit PID queue, waiting each PID before manifest upload and release registration, and adding regression assertions that ban jobs -pr/wait -n in this path. Tests run: bun run test tests/platform/customer-vps-host-bundle.test.ts; bash -n scripts/publish-release.sh; git diff --check; bun run check:patterns.

HamedMP:
[codex] Follow-up pushed for the latest Greptile review.\n- Replaced jobs -pr/wait -n job counting with an explicit incremental_upload_pids queue.\n- Throttle now waits the first queued PID before launching more uploads.\n- Final barrier waits every queued PID before the manifest upload and platform release registration.\n- Added regression assertions banning jobs -pr/wait -n in this upload path.\nCommit: c58a115a98\nTests: bun run test tests/platform/customer-vps-host-bundle.test.ts; bash -n scripts/publish-release.sh; git diff --check; bun run check:patterns.

PR #607, opened 2026-06-23
docs: correct Clerk-in-Docker-dev claim and document keyless workarounds

DennisSchm:
## What
The onboarding guide stated the Docker image bakes Clerk at build time, so dev
needs no Clerk keys. That is true for the production images (`Dockerfile`,
`Dockerfile.platform`) but **not** for `Dockerfile.dev`, which bakes no key.
With no key, `@clerk/nextjs` enters Keyless mode and the shell redirect-loops
on `/sign-in`.

## Changes
- `docs/dev/onboarding.md`: remove the inaccurate claim; add a "Clerk in Docker
  dev" subsection with the two real options (bypass flags, or Clerk dev keys),
  including the `restart` vs recreate `env_file` gotcha.
- `.env.docker.example`: surface both options as commented blocks.

## Scope
Docs + example-env only. No backend/code changes — Invariants section N/A.
`check:patterns` reports no TypeScript changes; typecheck/test/react-doctor N/A.

## Test plan
- Verified `Dockerfile.dev` contains no Clerk ARG/ENV (git history: never present).
- Confirmed production images and host-bundle build bake `NEXT_PUBLIC_CLERK_PUBLISHABLE_KEY`.
- Followed the documented bypass steps on a fresh `bun run docker`: desktop reachable, no loop.


HamedMP:

Status: conflicts with `main`; Docker development, Clerk fallback behavior, and onboarding documentation have moved since this external docs head.

To revive: verify every claim against the current keyless/local setup, rebase only the still-accurate public-safe guidance, and update current onboarding docs without restoring removed site content. Owner keep/kill decision requested.

PR #612, opened 2026-06-23
fix(terminal): align zellij close fallback

Nima-Naderi:
## Summary

- Add protocol support for `{ "type": "destroy" }` on canonical zellij `/ws/terminal/session` sockets.
- Route zellij WebSocket destroy through the canonical shell registry delete path with `{ force: true }`, while keeping `detach` as attach-process-only cleanup.
- Treat already-deleted/missing sessions as successful explicit-close cleanup and send only generic WebSocket errors for unexpected destroy failures.

## Tests

- `pnpm test tests/gateway/terminal-zellij-ws.test.ts` (passes)
- `pnpm test tests/gateway/terminal-zellij-ws.test.ts tests/gateway/shell-routes.test.ts tests/gateway/shell-registry.test.ts tests/shell/terminal-app-component.test.tsx tests/desktop/shell-socket.test.ts` (passes)
- `pnpm run typecheck:build-kernel` (passes)
- Equivalent package typecheck through gateway passed before platform typecheck failed on missing `stripe` module in this worktree dependency setup.
- `pnpm run check:patterns` (passes with existing scanner warnings only)
- `pnpm test` (attempted; fails on unrelated repo/environment issues, including missing desktop deps, platform telemetry constants, socket listen `EPERM`, app-runtime build/process tests, and other existing broad-suite failures)

## Review/Monitoring

Ready for CI/Greptile. CI is expected to provide authoritative clean-environment signal; local broad suite is noisy in this sandbox/worktree, but focused terminal suites pass.

## Invariants

- Source of truth: zellij terminal session lifecycle remains owned by the gateway shell session registry and its canonical delete path.
- Lock/transaction scope: no database transaction is introduced; in-memory attach cleanup remains local to the WebSocket handler, while session deletion stays inside the registry mutation path.
- Acceptable orphan states: REST DELETE failure or racing WebSocket cleanup should not leave explicit close without a protocol-correct destroy fallback; already-missing sessions during fallback are treated as successful cleanup.
- Auth source of truth: existing terminal WebSocket and REST authentication remain authoritative; no auth bypass or client-trusted ownership change is introduced.
- Deferred scope: terminal UI polish, non-zellij legacy behavior changes, and broad session lifecycle refactors are out of scope.

### DIFF
--- a/packages/gateway/src/server.ts
+++ b/packages/gateway/src/server.ts
@@ -431,6 +431,105 @@ function dispatchNamedTerminalMessage(
   });
 }
 
+type NamedTerminalRouteSession = {
+  onMessage(raw: string): void | Promise<void>;
+  onClose(): void;
+};
+
+type NamedTerminalPendingInputQueue = {
+  enqueue(raw: string): boolean;
+  drain(callback: (raw: string) => void): void;
+  clear(): void;
+};
+
+function isNamedTerminalDestroyFrame(raw: string): boolean {
+  let parsed: unknown;
+  try {
+    parsed = JSON.parse(raw);
+  } catch (err: unknown) {
+    if (!(err instanceof SyntaxError)) {
+      console.warn("[shell] terminal message parse failed:", err instanceof Error ? err.message : String(err));
+    }
+    return false;
+  }
+  return typeof parsed === "object" && parsed !== null && "type" in parsed && parsed.type === "destroy";
+}
+
+export function createNamedTerminalRouteController(options: {
+  pendingInput?: NamedTerminalPendingInputQueue;
+  onBufferOverflow?: () => void;
+} = {}) {
+  let namedHandle: NamedTerminalRouteSession | null = null;
+  let socketClosed = false;
+  let pendingDestroyFrame: string | null = null;
+
+  const dispatchDestroyOrClose = (session: NamedTerminalRouteSession) => {
+    const destroyFrame = pendingDestroyFrame;
+    pendingDestroyFrame = null;
+    if (destroyFrame) {
+      dispatchNamedTerminalMessage(session, destroyFrame);
+      return;
+    }
+    session.onClose();
+  };
+
+  return {
+    isSocketClosed(): boolean {
+      return socketClosed;
+    },
+    clearPendingState(): void {
+      pendingDestroyFrame = null;
+      options.pendingInput?.clear();
+    },
+    onOpenResolved(session: NamedTerminalRouteSession, openOptions: { drainPendingInput: boolean }): void {
+      if (socketClosed) {
+        options.pendingInput?.clear();
+        dispatchDestroyOrClose(session);
+        return;
+      }
+
+      namedHandle = session;
+      if (pendingDestroyFrame) {
+        options.pendingInput?.clear();
+        dispatchDestroyOrClose(session);
+        return;
+      }
+
+      if (openOptions.drainPendingInput) {
+        options.pendingInput?.drain((raw) => {
+          dispatchNamedTerminalMessage(session, raw);
+        });
+      }
+    },
+    onMessage(raw: string, messageOptions: { onBufferOverflow?: () => void } = {}): boolean {
+      if (namedHandle) {
+        dispatchNamedTerminalMessage(namedHandle, raw);
+        return true;
+      }
+
+      if (isNamedTerminalDestroyFrame(raw)) {
+        pendingDestroyFrame = raw;
+        return true;
+      }
+
+      if (!options.pendingInput) {
+        return false;
+      }
+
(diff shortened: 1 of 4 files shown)


Nima-Naderi:
## Problem statement

`TerminalPane` currently sends `{ "type": "destroy" }` over the active terminal WebSocket during explicit pane/tab close cleanup when `shouldDestroyOnUnmount` says the pane is being closed. That is protocol-correct for the legacy `/ws/terminal` path, but canonical zellij sessions use `/ws/terminal/session`, whose client message schema currently accepts only `input`, `resize`, `detach`, and `ping`.

`TerminalApp` also fires the canonical REST cleanup (`DELETE /api/terminal/sessions/:name?force=1`) on explicit close, but that request is best-effort and can fail or race with pane unmount. In that case the WebSocket cleanup fallback is supposed to preserve explicit close semantics, yet the zellij WebSocket rejects `destroy` as an invalid message and only tears down the attach process. The user-visible risk is an explicitly closed zellij terminal session surviving as a detached/orphaned session.

The fix needs to make explicit close reliable while keeping plain unmount/tab switch behavior as detach/cache only, and it should use the existing zellij session deletion source of truth rather than adding a second cleanup mechanism.

Nima-Naderi:
## Plan to fix

1. Add a focused failing gateway regression in `tests/gateway/terminal-zellij-ws.test.ts`: after opening `/ws/terminal/session`, a `{ "type": "destroy" }` client frame should delete the attached session through the zellij session registry with `{ force: true }`, close the attach process/socket, and not emit an `invalid_message` error.

2. Make the zellij WebSocket protocol explicitly support a `destroy` control frame in `packages/gateway/src/shell/ws.ts`. The handler should route destroy through the same registry deletion path used by `DELETE /api/terminal/sessions/:name?force=1`, so the gateway has one canonical session cleanup source of truth. A race where REST already deleted the session should be treated as successful cleanup for this explicit-close fallback; unexpected delete failures should be logged server-side and surfaced only as a generic WebSocket error before closing.

3. Keep `detach` semantics unchanged: `detach` closes only the attach process and socket, while `destroy` means explicit close and removes the zellij session. The existing `TerminalPane` cleanup decision can continue sending `destroy` only when `shouldDestroyOnUnmount` marks pane/tab close, but I will verify whether any shell-side regression test is needed around this distinction.

4. Add or update focused shell tests only if the existing coverage does not already prove explicit close sets `shouldDestroyOnUnmount` and sends the destroy frame. Avoid broad UI refactors and leave terminal styling/polish untouched.

5. Validate with the narrow gateway test first, then run the relevant shell/gateway terminal tests. If any React `.tsx` file changes are required, run `npx react-doctor@latest shell` before committing per repo policy.

Nima-Naderi:
## Final validation

Changes:
- Added route-level named zellij terminal lifecycle handling so destroy intent is preserved separately from queued pre-open input.
- Applied the destroy-before-open behavior to both `/ws/terminal/session` and `/ws/terminal?session=...`.
- Preserved close-without-destroy behavior as detach-only cleanup; pre-open non-destroy messages on `/ws/terminal?session=...` remain ignored.
- Added focused route-race coverage for destroy sent before async attach/open resolves, followed by socket close.

Validation:
- Commit validated: `24b46684cdb066f746bfd6d73ed35762a8f484db`.
- `/home/nima/matrix-os/node_modules/.bin/vitest run tests/gateway/terminal-zellij-ws.test.ts tests/gateway/terminal-zellij-route-race.test.ts` passed: 18 tests.
- `/home/nima/matrix-os/node_modules/.bin/tsc --noEmit -p packages/gateway/tsconfig.json` passed.
- `./scripts/review/check-patterns.sh` passed: 0 violations, existing warnings only.
- `git diff --check origin/main...HEAD` passed.
- No React files changed; `npx react-doctor@latest shell` was not required.
- The worktree `node_modules` cache was removed before finalizing; PR diff is scoped to source/tests only.

CI / review:
- `gh pr checks 612 --required` reports no required checks for this branch.
- Visible PR checks pass for the branch; no full CI matrix is attached to the PR check list for this branch.
- Greptile reports Confidence Score 5/5 on commit `24b46684`: https://github.com/HamedMP/matrix-os/pull/612#issuecomment-4780743341

Residual risk:
- Local focused gateway validation covers the explicit zellij destroy fallback and the follow-up route race. GitHub does not attach a full required CI matrix for this branch. The PR is not merged.

HamedMP:

Closing as superseded. Current terminal close/detach behavior is implemented in the shared Zellij runtime and canonical-session registry, including bounded attach/retry and persistent-session semantics. See [PR #643](https://github.com/HamedMP/matrix-os/pull/643), [PR #990](https://github.com/HamedMP/matrix-os/pull/990), and [PR #1133](https://github.com/HamedMP/matrix-os/pull/1133).

PR #626, opened 2026-06-24
fix(platform): resolve internal integrations for VPS machines

DennisSchm:
## Summary
- resolve internal integration requests against current `user_machines` records when no legacy `containers` row exists
- keep the existing legacy container lookup behavior
- add regression coverage for internal integration routing from a running VPS machine record
- add negative coverage for unknown handles and non-running VPS machines

## Context
Symphony was reaching the platform-owned Linear bridge at `/internal/containers/:handle/integrations/call`, but the platform returned `404 {"error":"Unknown handle"}` before Linear/Pipedream was reached. The internal integration middleware only checked the legacy `containers` table, while current customer VPS instances are represented in `user_machines`.

## Validation
- `/opt/matrix/runtime/node/bin/corepack pnpm --filter @matrix-os/observability build && /opt/matrix/runtime/node/bin/corepack pnpm --filter @matrix-os/kernel build && /opt/matrix/runtime/node/bin/corepack pnpm exec vitest run tests/platform/internal-integration-routes.test.ts`
- `tests/platform/internal-integration-routes.test.ts`: 5 tests passed

### DIFF
--- a/packages/platform/src/main.ts
+++ b/packages/platform/src/main.ts
@@ -3584,12 +3584,16 @@ export function createApp(deps: {
       }
 
       const record = await getContainer(db, handle);
-      if (!record?.clerkUserId) {
+      const runningMachine = record?.clerkUserId
+        ? undefined
+        : await getRunningUserMachineByHandle(db, handle);
+      const clerkUserId = record?.clerkUserId ?? runningMachine?.clerkUserId;
+      if (!clerkUserId) {
         return c.json({ error: 'Unknown handle' }, 404);
       }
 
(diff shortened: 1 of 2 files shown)


HamedMP:

Status: conflicts with `main`; platform startup and integration routes were subsequently split out of the old `main.ts` composition file.

To revive: reproduce the current VPS integration-resolution gap, add a route-level regression against the extracted platform modules, keep provider errors generic, and implement the smallest fix in the current route/service owner. Owner keep/kill decision requested.

PR #627, opened 2026-06-24
fix(terminal): hide zellij chrome in shell layouts

HamedMP:
## Summary
- Remove Zellij compact-bar plugin rows from Matrix-generated default shell layouts.
- Remove the same chrome from command-session and agent-runtime layouts so browser refreshes show only shell/program output.
- Update gateway regression tests to assert no compact/status/tab bar is generated.

## Tests
- bun run test tests/gateway/shell-zellij.test.ts tests/gateway/zellij-runtime.test.ts
- bun run check:patterns
- bun run typecheck
- git diff --check

## Invariants
- Source of truth: Matrix-generated Zellij layouts in gateway shell config, command-session layout generation, and zellij runtime layout generation are canonical for this terminal chrome behavior.
- Lock/transaction scope: no database or multi-write persistence path is changed; existing atomic config writes remain unchanged.
- Acceptable orphan states: already-running Zellij sessions may keep their old layout until recreated or refreshed through the normal generated config path.
- Auth source of truth: unchanged; terminal HTTP and websocket auth/session validation are not touched.
- Deferred scope: this does not change retained PTY lifecycle, attach semantics, release publishing, or live VPS rollout.


HamedMP:
Screenshot evidence for the post-removal terminal surface:

Chrome-free terminal shell

This shows the intended result after removing Zellij's compact/status/tab chrome: the terminal area displays only prompt and command output, with no transient "waiting to run ..." line.

PR #629, opened 2026-06-24
fix(terminal): soft detach zellij tabs

HamedMP:
## Summary

- Soft-detach Zellij attach clients on terminal WebSocket close/detach instead of killing them immediately.
- Keep hard-kill behavior as the fallback for attach processes that do not support soft detach.
- Add regression tests for named Zellij WebSocket close and Zellij adapter detach fallback behavior.

## Validation

- `pnpm exec vitest run tests/gateway/terminal-zellij-ws.test.ts`
- `pnpm exec vitest run tests/gateway/shell-zellij.test.ts`
- `bun run check:patterns`
- `pnpm --filter '@matrix-os/gateway' exec tsc --noEmit`
- `git diff --check`

## Invariants

**Source of truth**: Zellij remains the source of truth for named shell sessions and their panes; the WebSocket attach process is only a client view into that session.

**Lock/transaction scope**: No database or file writes are introduced. The critical lifecycle boundary is the WebSocket cleanup path, which now chooses `detach()` for soft client cleanup and only falls back to `kill()` when no detach primitive exists.

**Acceptable orphan states**: A named Zellij session may have zero attached WebSocket clients after tab switch, browser close, or network close. That is intentional; running panes remain available for later reattach. If the soft-detached attach client does not exit promptly, the adapter kills only that attach client after a bounded fallback timer.

**Auth source of truth**: WebSocket auth is unchanged and continues to use the existing `/ws/terminal` middleware/token path.

**Deferred scope**: This PR does not change session creation, session deletion, placement state, shell UI rendering, or legacy UUID PTY session TTL/eviction behavior.

### DIFF
--- a/packages/gateway/src/shell/ws.ts
+++ b/packages/gateway/src/shell/ws.ts
@@ -214,9 +214,13 @@ export function createShellWsHandler(options: ShellWsHandlerOptions) {
         return;
       }
       closed = true;
-      abortController.abort();
       cleanupProcessListeners();
-      child.kill();
+      if (typeof child.detach === "function") {
+        child.detach();
+      } else {
+        abortController.abort();
+        child.kill();
+      }
     };
 
     return {
--- a/packages/gateway/src/shell/zellij.ts
+++ b/packages/gateway/src/shell/zellij.ts
@@ -481,16 +481,11 @@ export function createZellijAdapter(deps: ZellijAdapterDeps = {}): ZellijAdapter
         if (exited) {
           return;
         }
-        try {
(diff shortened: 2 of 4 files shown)


HamedMP:

Closing as superseded. Soft client detach is now part of the shared persistent-session runtime rather than a tab-specific fallback: current `shell/ws.ts` detaches clients without destroying the named Zellij session, and [PR #1133](https://github.com/HamedMP/matrix-os/pull/1133) renders those soft clients on the canonical grid. See also [PR #990](https://github.com/HamedMP/matrix-os/pull/990).

PR #636, opened 2026-06-24
fix(gateway): initialize scratch project git repos

Nima-Naderi:
## Summary

- Initialize scratch projects as Git repositories with a `main` default branch and an empty initial commit that ignores global GPG signing.
- Forward branch worktree creation requests with `createBranch: true`, while service-layer logic only creates fallback branches for local/scratch projects.
- Preserve GitHub/remote safety: existing local branches still work, existing remote branches are tracked, and missing remote branches do not silently create typo branches.
- Add gateway regression coverage for scratch Git initialization, initialization cleanup, route forwarding, immediate scratch-project branch worktree creation using real Git, and remote typo-branch behavior.

## Tests

- `pnpm test tests/gateway/project-manager.test.ts tests/gateway/worktree-manager.test.ts tests/gateway/workspace-routes.test.ts`
- `pnpm --filter @matrix-os/gateway exec tsc --noEmit`
- `pnpm run check:patterns` (0 violations; existing repo-wide warnings only)
- Earlier `pnpm test` full-suite run was attempted and stopped after unrelated failures in platform/sync/app-runtime areas unrelated to this gateway change.

## Review/Monitoring

- Greptile 3/5 findings addressed in `bad5ea7b`:
  - scratch empty commit now passes `-c commit.gpgSign=false`
  - fallback branch creation is scoped to local/scratch projects
- Latest checks and Greptile review are being monitored after the update.

## Invariants

- **Source of truth**: project metadata remains `~/projects/<slug>/config.json`; scratch repo bytes live in `~/projects/<slug>/repo`; worktree metadata remains under `~/projects/<slug>/worktrees/<wt>/.matrix/worktree.json`.
- **Lock/transaction scope**: scratch project creation stays inside the existing per-project lock. Directory creation, Git initialization, initial commit, and config write are serialized for the slug. Worktree creation remains serialized by the same project lock and decides fallback behavior from the persisted project config. There are no network calls in the scratch path.
- **Acceptable orphan states**: if scratch Git initialization or initial commit fails before config write, the partially created project directory is removed and the API returns a generic `scratch_init_failed` error. If remote branch lookup fails or a remote branch is absent, remote/GitHub projects use the existing non-fallback `git worktree add` path and return the existing generic checkout failure on Git failure.
- **Auth source of truth**: route auth and owner scope remain the existing workspace request principal and injected `getOwnerScope`; this change does not add or bypass auth.
- **Deferred scope**: no UI changes, no server route ordering changes, no shell React changes, no migration for already-created non-Git scratch project directories.

### DIFF
--- a/packages/gateway/src/project-manager.ts
+++ b/packages/gateway/src/project-manager.ts
@@ -211,6 +211,8 @@ export function createProjectManager(options: {
               "user.name=Matrix OS",
               "-c",
               "user.email=matrix-os@example.invalid",
+              "-c",
+              "commit.gpgSign=false",
               "commit",
               "--allow-empty",
               "-m",
--- a/packages/gateway/src/project-manager.ts
+++ b/packages/gateway/src/project-manager.ts
@@ -201,12 +201,40 @@ export function createProjectManager(options: {
           }
           const repoPath = join(targetProjectPath, "repo");
           await mkdir(repoPath, { recursive: true });
+          try {
+            await runCommand("git", ["init", "-b", "main"], {
+              cwd: repoPath,
+              timeout: DEFAULT_TIMEOUT_MS,
+            });
+            await runCommand("git", [
+              "-c",
+              "user.name=Matrix OS",
+              "-c",
+              "user.email=matrix-os@example.invalid",
+              "commit",
+              "--allow-empty",
+              "-m",
+              "chore: initialize scratch project",
+            ], {
+              cwd: repoPath,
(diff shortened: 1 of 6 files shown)


HamedMP:

Status: conflicts with `main`; project/worktree creation and workspace routes have changed substantially.

To revive: reproduce scratch-project creation on the current project manager, define idempotent exclusive Git initialization, add current route/service tests first, and keep filesystem resolution inside the owner workspace boundary. Owner keep/kill decision requested.

PR #647, opened 2026-06-25
ci(shell): require production shell build

HamedMP:
## Summary

- Adds a canonical root `build:shell:production` script for the shell production build.
- Changes `shell/package.json` to run `next build --webpack`, matching `Dockerfile.platform` and `scripts/build-host-bundle.sh`.
- Adds a blocking `Shell Production Build` CI job after Type Check and before Pattern Scan, and wires it into `CI Results`.
- Uses only CI-safe public env vars for the shell build and does not set `POSTHOG_API_KEY` or `POSTHOG_PROJECT_ID`, so CI does not upload sourcemaps.

## Tests

- `pnpm install --frozen-lockfile`
- `CI=1 NEXT_TELEMETRY_DISABLED=1 NEXT_PUBLIC_CLERK_PUBLISHABLE_KEY=pk_test_Y2ktc2FmZS5leGFtcGxlLmNvbSQ= NEXT_PUBLIC_CLERK_SIGN_IN_URL=/sign-in NEXT_PUBLIC_CLERK_SIGN_UP_URL=/sign-up NEXT_PUBLIC_POSTHOG_KEY=phc_ci_shell_build NEXT_PUBLIC_POSTHOG_PROJECT_TOKEN=phc_ci_shell_build NEXT_PUBLIC_POSTHOG_HOST=https://eu.i.posthog.com NEXT_PUBLIC_POSTHOG_API_HOST=/relay bun run build:shell:production`
  - Expected guard behavior observed: fails on current `origin/main` shell production typecheck at `shell/src/hooks/useMatrixBillingAccess.ts:223`.
- `bun run check:patterns`
- `node -e 'const fs = require("node:fs"); const { parseDocument } = require("yaml"); const doc = parseDocument(fs.readFileSync(".github/workflows/ci.yml", "utf8")); if (doc.errors.length) { for (const err of doc.errors) console.error(err.message); process.exit(1); } console.log("workflow yaml ok");'`
- `git diff --check`
- `bun run typecheck`

## Review/Monitoring

- This PR intentionally makes shell production build failures blocking through `CI Results`.
- Current local production shell build fails on a pre-existing shell type error, which is the class of failure this guard is meant to catch before merge.
- Greptile and CI should be checked on the current PR head before merge; do not merge until the shell production build is green.

## Invariants

- Source of truth: CI configuration and package scripts only; no backend or data source-of-truth changes.
- Lock/transaction scope: not applicable; no runtime writes, database transactions, or locks are introduced.
- Acceptable orphan states: not applicable; the CI job either passes or blocks merge, and no persistent runtime state is created.
- Auth source of truth: unchanged; the build uses public compile-time env placeholders only and no private provider credentials.
- Deferred scope: fixing the currently surfaced shell type error is intentionally outside this PR's scoped files.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -116,9 +116,50 @@ jobs:
         if: needs.changes.outputs.should_run == 'true'
         run: bun run typecheck
 
+  shell-production-build:
+    name: Shell Production Build
+    needs: [changes, typecheck]
+    runs-on: ubuntu-latest
+    timeout-minutes: 15
+    steps:
+      - name: Skip expensive CI for docs-only changes
+        if: needs.changes.outputs.should_run != 'true'
+        run: echo "No CI-relevant source changes detected; skipping shell production build."
+
+      - uses: actions/checkout@v6
+        if: needs.changes.outputs.should_run == 'true'
+
+      - uses: oven-sh/setup-bun@v2
+        if: needs.changes.outputs.should_run == 'true'
+
+      - uses: pnpm/action-setup@v6
+        if: needs.changes.outputs.should_run == 'true'
+
+      - uses: actions/setup-node@v6
+        if: needs.changes.outputs.should_run == 'true'
+        with:
+          node-version: 24
+          cache: pnpm
+
+      - run: pnpm install --frozen-lockfile
+        if: needs.changes.outputs.should_run == 'true'
+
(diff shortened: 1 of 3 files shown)


Nima-Naderi:
Reopened under Nima-Naderi as #650: https://github.com/HamedMP/matrix-os/pull/650

PR #675, opened 2026-06-29
fix(vps): make code-server installer root-aware

HamedMP:
## Summary
- make matrix-install-tool-pack root-aware for code-server filesystem operations
- remove the code-service dependency cycle with matrix-code-server.service
- make the installer unit skip once code-server is already present

## Live root cause
On khorma9, matrix-code-server.service ran as root but install_code_server() invoked nested sudo for rm/mv/chown/chmod. Fresh VPS root had an expired password state, so sudo failed before code-server could be installed. After manual install, ExecStartPost=systemctl start matrix-code.service deadlocked with matrix-code.service Wants/After=matrix-code-server.service.

## Invariants
- Source of truth: /opt/matrix/runtime/code-server/bin/code-server is the installed runtime marker.
- Lock/transaction scope: existing /tmp/matrix-tool-pack-code-server.lock remains the only installer concurrency guard.
- Acceptable orphan states: matrix-code.service may start before the background installer and self-heal by invoking the same locked installer path.
- Auth source of truth: unchanged; code proxy auth stays in matrix-code and platform edge routing.
- Deferred scope: no broader developer-tool installer redesign in this patch.

## Tests
- bun run test tests/platform/customer-vps-host-bundle.test.ts tests/platform/customer-vps-cloud-init.test.ts tests/deploy/customer-vps/symphony-systemd.test.ts

### DIFF
--- a/distro/customer-vps/cloud-init.yaml
+++ b/distro/customer-vps/cloud-init.yaml
@@ -122,8 +122,7 @@ write_files:
     content: |
       [Unit]
       Description=Matrix OS customer code editor
-      After=matrix-restore.service matrix-code-server.service
-      Wants=matrix-code-server.service
+      After=matrix-restore.service
       Requires=matrix-restore.service
       ConditionPathExists=/opt/matrix/restore-complete
       ConditionPathExists=/opt/matrix/bin/matrix-code
@@ -151,6 +150,7 @@ write_files:
       Requires=matrix-restore.service
       ConditionPathExists=/opt/matrix/restore-complete
       ConditionPathExists=/opt/matrix/bin/matrix-install-tool-pack
+      ConditionPathExists=!/opt/matrix/runtime/code-server/bin/code-server
 
       [Service]
       Type=oneshot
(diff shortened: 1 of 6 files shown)


HamedMP:
[codex] Accepting the Greptile finding. I will align install_code_server() with the systemd guard by checking /opt/matrix/runtime/code-server/bin/code-server instead of command -v code-server, and add regression coverage so a leftover shim cannot mask a missing runtime binary.

HamedMP:
[codex] Addressed Greptile P2 in 4dec9aa9ad: install_code_server() now checks the canonical runtime binary at `/code-server/bin/code-server`, matching the systemd `ConditionPathExists` guard. Added host-bundle assertions so a leftover shim cannot mask a missing runtime binary. Tests: `bun run test tests/platform/customer-vps-host-bundle.test.ts tests/platform/customer-vps-cloud-init.test.ts tests/deploy/customer-vps/symphony-systemd.test.ts` (80/80).

PR #683, opened 2026-06-30
docs(self-host): add web credential rotation

HamedMP:
## Summary
- Add self-host docs for changing the nginx Basic Auth web username and password
- Clarify that web credentials are separate from the Linux matrix user, CLI token, Postgres, and Matrix Cloud login
- Include a verification curl command for the new credentials

## Tests
- git diff --check
- Not run: bun run test tests/scripts/self-host-installer.test.ts (fresh worktree has no node_modules; tsc not found)

## Invariants
- Source of truth: nginx Basic Auth remains stored in /etc/nginx/matrix-self-host.htpasswd.
- Lock/transaction scope: docs-only change; no runtime or installer behavior changes.
- Acceptable orphan states: /opt/matrix/env/initial-ui-user is an operator note only; nginx continues to read the htpasswd file.
- Auth source of truth: browser edge auth remains nginx Basic Auth; CLI auth remains MATRIX_AUTH_TOKEN.
- Deferred scope: no change to standalone mobile/desktop login or managed Matrix Cloud auth.

### DIFF
--- a/www/content/docs/self-host.mdx
+++ b/www/content/docs/self-host.mdx
@@ -178,11 +178,55 @@ The preview installer is intentionally minimal: nginx Basic Auth protects the br
 
 - Put the host behind HTTPS with DNS, Tailscale, Cloudflare Access/Tunnel, or another trusted edge.
 - Keep ports `3000`, `4000`, `8787`, `8788`, and `5432` closed to the public internet.
-- Rotate `/opt/matrix/env/initial-ui-password` and `MATRIX_AUTH_TOKEN` if either is exposed.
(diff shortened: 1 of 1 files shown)


HamedMP:

Status: conflicts with `main` and targets the removed `www` tree. Credential rotation remains security-sensitive and must match the current self-hosted auth/config surface.

To revive: verify the current rotation procedure end to end, keep examples secret-free, place public documentation in the separate docs site, and add a current repository contract test if code/config behavior is claimed. Owner keep/kill decision requested.

Issue #966, opened 2026-07-13
P2 follow-ups from reland review (#964)


HamedMP:
Greptile P2 findings on the reland PR #964 (content previously reviewed 5/5 as #934/#933/#926), deferred per review policy:

1. `packages/gateway/src/coding-agents/routes.ts:442` — `deps.projectMutations!` non-null assertion inside an async closure despite the surrounding guard; bind to a local const before the handler.
2. `apps/mobile/components/AgentComposerScreen.tsx:420` — project picker stays open after successful project creation; should close/advance on success.
3. `apps/mobile/lib/agent-cockpit.ts:49` — `threadGroup` switch lacks an exhaustive default for future attention values; add a default returning "recent" (or exhaustiveness check).

PR #970, opened 2026-07-13
fix(mobile): files browser review hardening

HamedMP:
Lands the review-fix commits for the files browser (content itself reached main via the #953 merge ancestry; this PR carries only the fixes addressing #948's findings):

- `96058b2ff` guard files screen against stale responses (generation guards)
- `453af9f1f` bound file preview downloads (no download without Content-Length; RN fetch has no body streams)
- `4b9a8f0c8` invalidate stale file responses at event time (guards bump in event handlers, not effects)

Original review context: #948 (closed in the merge cascade).


HamedMP:
Deferred: the chunked/no-Content-Length P2 stays as designed — RN fetch exposes no body streams, so downloading without a length header reintroduces the unbounded-buffer security finding this PR fixes. Preview for such responses intentionally reports unavailable; revisit if the gateway adds explicit lengths or a ranged read. Tracked with the other deferrals in #966.

PR #1024, opened 2026-07-18
fix(docker): disable global virtual store inside image builds

HamedMP:
## Summary

Platform Cloud Run has been red on main since this morning: the platform image build fails in `pnpm --dir shell exec next build --webpack` with import errors such as "'Emitter' is not exported from '../utils/event.js'" (langium via mermaid/streamdown).

Root cause: #1008 enabled `enableGlobalVirtualStore` in `pnpm-workspace.yaml`. Inside Docker image builds, node_modules then symlinks into the build stage's home directory (`/root/.local/share/pnpm/store/v10/links/...`), which breaks webpack's module resolution for package-internal relative imports and would also leave dangling symlinks when node_modules is copied into later image stages.

Fix: pass `--config.enableGlobalVirtualStore=false` to every `pnpm install` in `Dockerfile` and `Dockerfile.platform`, keeping the global store benefit for local worktrees while image builds stay self-contained.

## Validation

- Reproduced the failure from a clean `git archive` of main: `docker build -f Dockerfile.platform` fails at the shell webpack build with the same errors as Cloud Build.
- Rebuilt the same clean tree with this patch: build passes the previously failing step (full-build evidence in the PR checks and local log).

## Invariants

- Source of truth: `pnpm-workspace.yaml` remains the repo-wide pnpm configuration; image builds override only the virtual-store location at install time.
- Lock/transaction scope: none; no lockfile or dependency version changes.
- Acceptable orphan states: none; the flag only changes where installed packages physically live inside build stages.
- Auth source of truth: unchanged.
- Deferred scope: host-bundle and CI-runner installs keep the global store (they build on the runner filesystem and are unaffected).

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #1028, opened 2026-07-18
fix(docker): isolate dev container dependency links

Nima-Naderi:
## Summary

- preserve `enableGlobalVirtualStore: true` for host worktrees
- force both Docker dev entrypoint install paths to use a container-local virtual store
- cover the host/container split and every entrypoint install call with a regression test
- complement #1024, which scopes image-build installs; this PR fixes the `Dockerfile.dev` runtime entrypoint and its root-to-`matrixos` user switch

The Docker dependency volume must remain self-contained. The entrypoint installs dependencies as root and then launches services as `matrixos`; global-store links under root's data directory are unreadable after that switch, so TypeScript reports widespread module-not-found errors and the Docker CI health check never becomes ready.

## Tests

- `pnpm exec vitest run tests/scripts/docker-dev-entrypoint.test.ts` — pass (2 tests)
- `bash -n distro/docker-dev-entrypoint.sh` — pass
- `pnpm run check:patterns` — pass (0 violations; existing warnings only)
- TypeScript projects — observability, kernel, gateway, platform, proxy, edge-router pass; unchanged desktop files have two `origin/main` implicit-`any` errors
- `pnpm run test` — 8,648 pass / 20 skip; seven unchanged UI suites fail to load because the host global-store layout cannot resolve `vitest` from `@testing-library/jest-dom`; the added regression passes

No React files or user-visible UI changed, so React Doctor and screenshot evidence do not apply.

### Source of truth

`pnpm-workspace.yaml` remains the canonical host-worktree setting. The scoped CLI flag in `distro/docker-dev-entrypoint.sh` is the canonical exception for the Docker dev dependency volume.

### Lock/transaction scope

No database writes, locks, or application network calls change. Initial dependency installation and lockfile-triggered reinstallation use the same local virtual-store layout.

### Acceptable orphan states

Install failures keep the existing retry/failure behavior. A volume created with global-store links lacks the local `.pnpm` sentinel, so the entrypoint reinstalls it into the self-contained layout before starting services.

### Auth source of truth

Unchanged. Root performs setup, then the entrypoint retains the existing `matrixos` runtime-user boundary; no credentials or environment variables are added.

### Deferred scope

- production/image-build portability remains covered by #1024
- no attempt is made to share the global virtual store from inside containers
- no change is made to the host global-store setting

### DIFF
--- a/distro/docker-dev-entrypoint.sh
+++ b/distro/docker-dev-entrypoint.sh
@@ -16,7 +16,10 @@ ensure_deps() {
     return 0
   fi
   echo "[matrix-os-dev] Installing dependencies (lockfile changed)..."
-  pnpm install --frozen-lockfile
+  # Host worktrees share pnpm's global virtual store, but this container
+  # installs as root before dropping to matrixos. Keep the Docker dependency
+  # volume self-contained so its links remain readable after the user switch.
+  pnpm install --frozen-lockfile --config.enableGlobalVirtualStore=false
   md5sum pnpm-lock.yaml > node_modules/.pnpm-lock-hash 2>/dev/null || true
 }
 ensure_deps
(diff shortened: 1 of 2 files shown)


HamedMP:
### Merge activity

* **Jul 18, 3:43 PM UTC**: A user started a stack merge that includes this pull request via [Graphite](https://app.graphite.com/github/pr/HamedMP/matrix-os/1028).
* **Jul 18, 3:44 PM UTC**: @HamedMP merged this pull request with [Graphite](https://app.graphite.com/github/pr/HamedMP/matrix-os/1028).

HamedMP:
* **#1028** <a href="https://app.graphite.com/github/pr/HamedMP/matrix-os/1028?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/HamedMP/matrix-os/1028?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

PR #1008, opened 2026-07-16
chore(pnpm): enable global virtual store

HamedMP:
## Summary

- enable pnpm's dependency-graph-addressed global virtual store for all workspace installs
- reduce per-worktree node_modules overhead while preserving the shared content-addressable store

## Validation

- pnpm config get enableGlobalVirtualStore returns true
- pnpm install --frozen-lockfile passed in a fresh detached disposable worktree
- 3,019 packages reused with zero downloads
- disposable worktree node_modules used 13 MB locally and linked into the central pnpm store

## Invariants

- Source of truth: pnpm-workspace.yaml
- Lock/transaction scope: no database or runtime lock scope; pnpm-lock.yaml is unchanged
- Acceptable orphan states: existing worktrees keep their current node_modules until reinstalled; new installs use the global virtual store
- Auth source of truth: unchanged; this configuration does not affect authentication
- Deferred scope: bulk reinstalling active or dirty worktrees is intentionally excluded to avoid disrupting in-progress work

PR #1029, opened 2026-07-18
fix(docker): keep dev container installs on the local virtual store

HamedMP:
## Summary

Follow-up to #1024. Docker scenario tests on main fail with `Cannot find module '/app/packages/kernel/node_modules/typescript/bin/tsc'`: the dev container entrypoint installs dependencies as root, but services and package builds run as the `matrixos` user. With the repo-wide `enableGlobalVirtualStore` (#1008), node_modules symlinks into /root's pnpm store, which the service user cannot resolve.

Fix: pass `--config.enableGlobalVirtualStore=false` to both `pnpm install` calls in `distro/docker-dev-entrypoint.sh` (initial install and the lockfile-watcher reinstall), keeping the dev container's virtual store inside the repo.

## Validation

- `bash -n` on the entrypoint.
- Same flag already proven in #1024 for the platform and user image builds (Platform Cloud Run green on main after merge).

## Invariants

- Source of truth: `pnpm-workspace.yaml` remains the repo-wide pnpm config; only containerized installs override the store location.
- Lock/transaction scope: none; no dependency changes.
- Acceptable orphan states: none.
- Auth source of truth: unchanged.
- Deferred scope: `scripts/setup-server.sh` installs and runs as the same user, so it keeps the global store.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/distro/docker-dev-entrypoint.sh
+++ b/distro/docker-dev-entrypoint.sh
@@ -16,7 +16,10 @@ ensure_deps() {
     return 0
   fi
   echo "[matrix-os-dev] Installing dependencies (lockfile changed)..."
-  pnpm install --frozen-lockfile
+  # Installs run as root but services run as matrixos; the repo-wide global
+  # virtual store would symlink node_modules into /root and break tool
+  # resolution (e.g. typescript/bin/tsc) for the service user.
+  pnpm install --frozen-lockfile --config.enableGlobalVirtualStore=false
   md5sum pnpm-lock.yaml > node_modules/.pnpm-lock-hash 2>/dev/null || true
 }
 ensure_deps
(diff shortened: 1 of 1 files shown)


HamedMP:
* **#1029** <a href="https://app.graphite.com/github/pr/HamedMP/matrix-os/1029?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/HamedMP/matrix-os/1029?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

HamedMP:
Superseded by #1028, which lands the same --config.enableGlobalVirtualStore=false fix on both dev-entrypoint installs plus a regression test.

PR #1075, opened 2026-07-20
feat(mobile): add email-code sign-in and enable EAS Update

HamedMP:
## Why

The Matrix OS Clerk instance enables `email_code` as its only non-OAuth first
factor, so `<SignIn />` on the web offers Google, GitHub, **and** an email
one-time code. The native app only ever rendered the two OAuth buttons, so
anyone whose account is email-based could not sign in on mobile at all.

Separately, `expo-updates` was not installed. No shipped build could receive an
over-the-air JS update, so every fix required a full store release.

## What changed

**Email-code sign-in**
- `lib/clerk-sign-in.ts` drives the Clerk flow (`create` -> `prepareFirstFactor`
  -> `attemptFirstFactor` -> `setActive`) with no React or SDK imports, so the
  branching is unit testable.
- The hosted panel gained an email field and a code step, next to the existing
  OAuth buttons. Self-hosted computers are unchanged — they have no Clerk
  instance and keep basic auth.

**EAS Update (OTA)**
- Added `expo-updates`, pointed `app.json` at this project's EAS Update
  endpoint, set `runtimeVersion.policy` to `appVersion`, and gave every
  `eas.json` build profile its own channel.
- This build is the first that can receive OTA updates; builds made before it
  cannot, because they ship without the native module.

**Refactor**
- `SignInScreen` was 285 lines before this change and the email flow pushed it
  over react-doctor's 300-line component limit. Split into `GatewayUrlPanel`,
  `HostedSignInPanel`, `EmailCodeForm`, and `SelfHostedSignInPanel`. No
  behaviour change to the existing panels.

**Mobile test gate repair (incidental but required)**
`pnpm --dir apps/mobile exec jest` could not start at all on this branch's base:
the workspace shares a global pnpm virtual store, so packages are linked from
content-addressed directories outside the repo and cannot resolve dependencies
they never declared. Fixed with the repo's existing mechanisms — two
`packageExtensions` entries for `require()`-time loads (jest-expo's preset,
react-native-worklets' babel plugin) and a `modulePaths` fallback in
`apps/mobile/jest.config.js` for everything Jest's own resolver loads.

## Invariants

- **Source of truth**: Clerk owns the session. The app never stores the code or
  the pending attempt beyond the screen — the attempt lives in a ref and is
  cleared on success or on "Use a different email".
- **Auth source of truth**: `setActive({ session })` is the only way a session
  becomes active; a non-`complete` status or a missing `createdSessionId` is
  treated as failure, never as success.
- **Acceptable orphan states**: if `create` succeeds but `prepareFirstFactor`
  fails, the attempt is discarded and the user stays on the email step. Clerk
  expires the orphaned attempt on its own.
- **Error policy**: only Clerk's own user-facing `longMessage`/`message` is
  rendered. Network failures, thrown non-Clerk values, and SDK internals collapse
  to a fixed generic string, so no raw internals reach the screen.
- **OTA safety**: `runtimeVersion.policy: "appVersion"` means an update only
  reaches builds with a matching app version. A release that changes native code
  must bump `version` in `app.json`, or the update will land on mismatched
  native code.
- **Deferred scope**: sign-**up** by email is not in scope — the instance
  requires a username and legal consent at sign-up, which is a separate flow.
  Mobile sign-up still goes through OAuth or the web.

## Verification

- `pnpm --dir apps/mobile exec jest --runInBand` — 56 suites / 600 tests pass.
  Previously the suite could not start.
- `npx react-doctor@latest --scope changed` — 100/100, no issues.
- `bun run check:patterns` — 0 violations.
- `pnpm --filter matrix-os-mobile exec tsc --noEmit` — 31 errors, byte-identical
  to `origin/main` with base dependencies installed (verified by reinstalling the
  base `package.json` files and re-running). All are the pre-existing
  `'X' cannot be used as a JSX component` class-component/React 19 typing issue
  in untouched files; none originate in this diff.


HamedMP:
## Verification

**Screenshot (iOS 26.5 simulator, iPhone 17 Pro, dev build of this branch)**

The hosted panel now renders Google, GitHub, an `or` divider, the email field, and `Email me a code`. Self-hosted computers are unchanged and still show Basic Auth.

**Keyboard fix**: the reporter hit the email input being hidden behind the keyboard. The centred layout meant iOS `padding` avoidance only shrank the viewport rather than lifting the focused field; the scroll view now adjusts its own keyboard insets. Covered by a regression test in `sign-in-screen.test.tsx`.

**Gates on this head**
- `pnpm --dir apps/mobile exec jest --runInBand` — 56 suites / 606 tests pass
- `npx react-doctor@latest --scope changed` — 100/100, no issues
- `bun run check:patterns` — 0 violations
- `pnpm --filter matrix-os-mobile exec tsc --noEmit` — 31 errors, identical to `origin/main` with base dependencies (pre-existing React 19 class-component JSX typing, none from this diff)

**OTA wiring confirmed natively**: after prebuild, `ios/MatrixOS/Supporting/Expo.plist` contains `EXUpdatesEnabled=true`, `EXUpdatesURL=https://u.expo.dev/cdd27a8d-…`, `EXUpdatesRuntimeVersion=0.2.1`, and `EXUpdates 57.0.8` is in `Podfile.lock`. The running dev build reports `Runtime version: 0.2.1`.

App version bumped `0.2.0` → `0.2.1` so this TestFlight build is distinct from the existing 0.2.0 build (which cannot receive OTA updates, having shipped without the native module).

PR #1124, opened 2026-07-28
fix(billing): recover unavailable checkout sessions

HamedMP:
## Summary

- send the direct allowlisted CLI device return path to checkout
- let users resume an already-open Stripe Checkout instead of showing a generic outage
- require checkout.session.completed and checkout.session.expired on the production Stripe webhook before deployment

## Root cause

The per-runtime active checkout lock shipped on July 21, but the production Stripe endpoint subscribed only to subscription events. Completed and expired Checkout Sessions therefore remained open in platform Postgres and later attempts returned 409. The CLI billing wrapper also produced a returnPath rejected by the platform allowlist.

## Validation

- 109 targeted billing, shell, and workflow tests pass
- platform/gateway TypeScript checks pass
- BillingPanel ESLint passes
- git diff --check passes

## Validation limitations

- repo-wide typecheck reaches an unrelated desktop Vite 6/7 type mismatch when this isolated worktree borrows the root install
- BillingGate already has a React ref lint violation at line 513 outside the changed code

## Production follow-up

After approval, add the two lifecycle events to the exact production billing webhook and reconcile currently open checkout-attempt rows against Stripe session truth.


HamedMP:
Addressed all Greptile findings on head `0b49e53ce7`:

- mismatched selections now return bounded selection metadata, never the stale Stripe URL; the UI names the original plan/interval/region so the user can safely re-select it
- Stripe webhook discovery now follows `has_more` / `starting_after` pagination
- added and verified a production-shell Playwright visual regression for the conflict state

Billing checkout selection conflict

Validation: 109/109 focused tests pass, production shell build passes, and the non-updating Playwright screenshot assertion passes.

HamedMP:
@greptileai review

HamedMP:
Addressed the current-head Greptile legacy-state finding in `dd1e62a2c3`.

An open attempt with pre-migration null selection fields is now retired only after the 24-hour maximum Stripe Checkout lifetime has elapsed. The conditional retirement and replacement claim execute in one transaction; younger sessions remain protected because they may still be payable.

Added a regression that starts from a legacy open row, verifies the old row becomes `abandoned`, and verifies the replacement is claimed with the requested selection. Validation: 110/110 focused billing tests pass, platform typecheck passes, and `git diff --check` passes.

HamedMP:
@greptileai review

HamedMP:
Addressed the younger-legacy-session blocker on head `825b01396b` by replacing the age-only assumption with bounded, authoritative Stripe reconciliation.

For an open legacy DB row with missing selection fields, the platform now retrieves that exact Checkout Session and validates ownership. An exact provider price/region match safely resumes its URL; a mismatch returns bounded selection metadata without exposing the URL; an expired provider session is resolved and replaced; a completed session remains in payment-settling state. Stripe errors remain server-only and map to the existing generic unavailable response.

Validation: 127/127 focused billing/workflow tests pass, platform typecheck passes, and `git diff --check` passes.

HamedMP:
@greptileai review

PR #1132, opened 2026-07-31
fix(terminal): preserve ANSI contrast in light themes

Nima-Naderi:
## Summary

- Correct Paper Light ANSI color 0 from the background value `#FBF1C7` to
  Gruvbox dark0 `#282828`.
- Apply xterm's native `minimumContrastRatio: 4.5` policy to light terminal
  backgrounds at creation time and on live terminal-theme changes.
- Add parser, theme-invariant, renderer-path, selection, cursor, reverse-video,
  reset, prompt, and explicit-black regression coverage.

The implementation uses option C: a semantic palette correction plus bounded
contrast enforcement. Option A alone fixes color 0 but leaves other light-theme
palette entries unprotected. Option B alone masks the invalid color-0 palette
assignment instead of correcting it. Dark terminal themes retain xterm's default
ratio of `1`.

## Root Cause

The exact Paper Light theme assigned both ANSI black/color 0 and the terminal
background to `#FBF1C7`. xterm correctly parsed SGR 30 and `38;5;0` as palette
index 0, so their effective foreground was pixel-identical to the background
before this change.

This proves a terminal-theme defect. It does **not** prove that the reported
original `ls` incident used ANSI black; the representative shipped `ls` path
described below did not emit it.

## Evidence

Exact reproduction input under `getTerminalThemePreset("light")`:

```text
default \x1b[30mSGR30\x1b[0m \x1b[38;5;0mIDX0\x1b[0m
```

Before the fix, xterm 6 parsed both colored cells as palette foreground index
`0`, and the effective colors were:

```text
background = #FBF1C7
ANSI black = #FBF1C7
collision  = true
```

After the fix, Paper Light color 0 is `#282828`, with a raw contrast ratio of
`12.9942:1` against `#FBF1C7`. Light terminals additionally request xterm's
`4.5:1` minimum contrast handling.

The shipped gateway terminal environment and generated shell wrapper were run
under a real PTY with `TERM=xterm-256color`, `COLORTERM=truecolor`, `CLICOLOR=1`,
`FORCE_COLOR=3`, and `COLORFGBG=15;0`. `LS_COLORS` was unset. The available
generated wrapper selected bash (zsh was not installed on the validation host):

- the Matrix prompt emitted bold cyan (`0;1;36`) and bold blue (`0;1;34`);
- plain `/usr/bin/ls -1` emitted no ANSI for files, directories, executables, or
  symlinks;
- `ls --color=auto` and `ls --color=always` emitted directories as bold ANSI
  blue (`01;34`), executables as bold green (`01;32`), symlinks as bold cyan
  (`01;36`), and regular files with the default foreground;
- no representative `ls` stream used ANSI black, 256-color, or truecolor.

Therefore the original `ls` stream was **not implicated** by this reproduction.

## Tests

- `bun run typecheck` — passed.
- `bun run check:patterns` — passed with zero violations; five unrelated
  repository-wide warning categories remain.
- Focused theme/component tests — 4 files, 42 tests passed.
- Previously failed full-run files — 2 files, 147 tests passed in isolation.
- `bun run test -- --maxWorkers=4` — 828 files and 9,215 tests passed; the
  shared-host run exited nonzero after stale `/tmp/test-home` state contaminated
  `session-registry.test.ts` and one unrelated TerminalApp test exceeded its
  20-second timeout. After quarantining that test-owned temp state, both complete
  files passed in isolation (147/147). An earlier unbounded `bun run test` also
  encountered global-store peer resolution and worker-exit failures; the local
  pnpm virtual-store relink fixed the peer-resolution failures.
- `npx react-doctor@latest shell --verbose --scope changed` — no finding caused
  by this diff. The reported existing `TerminalPane` observer finding is a false
  positive; the same effect cleanup calls `resizeObserver.disconnect()`.

## Manual Validation

Rendered an ANSI matrix in a real browser with xterm's default renderer and the
WebGL addon side by side. Both used the exact Paper Light theme and a `4.5`
minimum contrast ratio. The WebGL surface created three canvases and loaded its
addon; the default surface loaded no WebGL addon. Default foreground, prompts,
SGR 30, SGR 34, SGR 90, palette index 0, resets, reverse video, all standard and
bright colors, selection, and cursor remained visible.

Paper Light ANSI contrast in default and WebGL renderers

## Accessibility Impact

- Paper Light explicit black now has `12.9942:1` raw contrast.
- Every light preset receives xterm's WCAG AA `4.5:1` foreground floor for both
  normal and WebGL rendering paths.
- Tests preserve selection compositing, foreground/background reversal, cursor
  contrast, default foreground, blue, bright black, and simultaneous prompt and
  explicit-black visibility.
- Dark terminal themes and their palettes are unchanged.

## Invariants

- **Source of truth:** terminal palette and contrast policy remain in the
  terminal theme module and are applied per terminal instance.
- **Terminal locality:** no global shell-theme state is introduced or mutated.
- **Transport/lifecycle:** terminal sizing, Zellij ownership, WebSocket delivery,
  replay, and WebGL selection behavior are unchanged.
- **Deferred scope:** no `ls` special case and no claim that this resolves the
  original incident without reproducing that incident's exact ANSI stream.

## Review / Monitoring

- One explicit Greptile request was made after the non-draft PR content was
  verified. Greptile returned 5/5 on commit
  `2a81766b1be03a06cf265fca90cd99ff9f1845f5` with no concrete findings or
  review threads.
- `ready-for-ci` was added only after that trusted 5/5 review.
- All triggered CI passed: title, typecheck, pattern scan, React Doctor, docs
  contracts, shell production build, sync client package, all four unit-test
  shards, E2E, aggregate CI, Docker image build, and Docker smoke test.
- Preview jobs were skipped because this PR does not carry a preview label.
  The PR remains open and must not be merged by this workflow.


Nima-Naderi:
@greptileai please review this PR when ready.

PR #1143, opened 2026-08-03
fix(desktop): ship signed launchable macOS builds

HamedMP:
## Summary

- bundle `@matrix-os/contracts` into Electron runtime bundles and keep its raw TypeScript sources out of `app.asar`
- add the explicit `electron-vite -> electron` dependency edge required by pnpm's global virtual store
- make macOS certificate and notarization credentials mandatory for every release build
- verify Developer ID signing, stapling, Gatekeeper, ASAR contents, DMG mounting, and packaged-app startup before artifacts upload
- update the desktop release runbook and add regression coverage

## Root cause

The main-process build externalized the workspace contracts package. electron-builder then copied the package's raw `src/index.ts` into `app.asar/node_modules`, where packaged Electron refuses Node type stripping and crashes at startup. Separately, the release workflow explicitly disabled signing when Apple credentials were absent, allowing unsigned artifacts to be published.

## Tests

- `pnpm exec vitest run tests/desktop/desktop-release-workflows.test.ts tests/desktop/packaging.test.ts` — 18 passed
- `pnpm --filter desktop run typecheck`
- `pnpm --filter desktop build`
- `CI=true pnpm install --frozen-lockfile`
- `bun run check:patterns:diff` — 0 violations
- parsed `.github/workflows/desktop-build.yml` with the workspace YAML parser
- built an unsigned local arm64 directory package for inspection: raw contracts absent from ASAR
- launched the packaged arm64 app with isolated user data for 10 seconds: app remained running and the type-stripping crash did not recur

## Release follow-up

The repository currently has none of the required Apple release secrets. Before dispatching the next Desktop Canary release, add the certificate, certificate password, Apple ID, app-specific password, and team ID documented in `docs/dev/desktop-release.md`. The workflow now fails closed until they exist.

### Source of truth

- `desktop/out/**` is the executable Electron bundle; workspace TypeScript must not be a packaged runtime dependency.
- GitHub Actions secrets are the sole source for the Developer ID certificate and Apple notarization credentials.

### Lock/transaction scope

- No database, transaction, or shared lock behavior changes.
- Signing, notarization, package inspection, and launch verification complete before artifact upload.

### Acceptable orphan states

- Missing credentials or any verification failure leaves a failed workflow with no uploaded release artifact.
- DMG mounts, copied apps, and launch processes are cleaned up by the workflow trap.

### Auth source of truth

- GitHub Actions repository secrets are required; there is no unsigned fallback.

### Deferred scope

- Provisioning Apple credentials and dispatching the first signed canary require repository-owner access and are not encoded in this PR.
- An existing non-fatal updater initialization warning observed during local launch smoke is unrelated to the reported startup crash.

### DIFF
--- a/.github/workflows/desktop-build.yml
+++ b/.github/workflows/desktop-build.yml
@@ -66,38 +66,35 @@ jobs:
           MAC_CERTIFICATE_PASSWORD: ${{ secrets.MATRIX_DESKTOP_MAC_CERTIFICATE_PASSWORD || secrets.CSC_KEY_PASSWORD }}
         run: |
           set -euo pipefail
-          if [ -n "${MAC_CERTIFICATE:-}" ]; then
-            cert_delimiter="MATRIX_DESKTOP_CERT_$(uuidgen | tr '[:lower:]' '[:upper:]' | tr -d '-')"
-            password_delimiter="MATRIX_DESKTOP_CERT_PASSWORD_$(uuidgen | tr '[:lower:]' '[:upper:]' | tr -d '-')"
-            {
-              echo "CSC_LINK<<$cert_delimiter"
-              printf '%s\n' "$MAC_CERTIFICATE"
-              echo "$cert_delimiter"
-              echo "CSC_KEY_PASSWORD<<$password_delimiter"
-              printf '%s\n' "${MAC_CERTIFICATE_PASSWORD:-}"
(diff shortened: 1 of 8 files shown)


HamedMP:
@greptileai review current head a9e1238513ca2b162b54388e97836a793a0db1b3

PR #1156, opened 2026-08-05
fix(runtime): keep OpenClaw npm cache writable under systemd

xyhStruggler:
## Summary

- Redirect OpenClaw's npm cache to service-private temporary storage.
- Preserve the existing `ProtectHome=read-only` systemd hardening.
- Give runtime-switch mutations a dedicated 90-second Shell timeout while keeping ordinary agent-setting mutations at 15 seconds.
- Disable both provider Configure actions while an agent-settings mutation is pending.
- Add focused regressions for the hardened npm cache, client timeout contracts, and pending UI state.

## Root cause

OpenClaw repairs its missing Matrix plugin during gateway startup. npm defaults to `~/.npm`, but the customer VPS unit intentionally makes the owner home read-only. The write fails with `EROFS`, activation fails, and the runtime controller rolls back to Hermes.

The wrapper now exports `npm_config_cache` beneath the unit's writable `PrivateTmp` namespace. This keeps repair artifacts ephemeral and avoids expanding filesystem permissions or storing cache data in owner state.

Live preview testing exposed a second timing issue. The Shell stopped waiting for every agent-setting mutation after 15 seconds, while the VPS runtime controller permits a switch to run for 75 seconds. Aborting the browser request does not cancel the controller operation, so the UI could report a failure even though OpenClaw became healthy and active shortly afterward. A retry could then return a revision conflict.

Runtime-switch mutations now wait up to 90 seconds, past the controller deadline. Other agent-setting writes keep the existing 15-second bound. While that mutation is pending, runtime switching and both Configure actions share the same busy state so configuration cannot race the transition.

## User impact

Switching from Hermes to an installed OpenClaw runtime can complete when the Matrix plugin needs first-start repair. The runtime and Configure actions remain disabled while the request is pending, and the Shell no longer times out before the server's bounded transition deadline. Existing rollback behavior remains unchanged if activation fails.

## Tests

- TDD timeout regression failed with runtime switches still using 15 seconds, then passed with the dedicated 90-second bound.
- TDD UI regression failed while Configure remained enabled during a deferred runtime switch, then passed after binding both Configure actions to the shared busy state.
- Focused Shell/OpenClaw suites: 43/43 passed locally.
- `pnpm --filter shell exec tsc --noEmit`: passed locally.
- `./scripts/review/check-patterns.sh`: 0 violations; existing warnings only.
- `git diff --check`: passed.
- Current-head Greptile: 5/5 on `631248ce5407b9ec2848b19d4e26fac2226dbb3f`.
[...]

PR #1154, opened 2026-08-04
fix(agent): repair OpenClaw install and add Hermes settings UI

HamedMP:
## Summary

- Repair upgraded VPSes whose retained Node 24.14.1 runtime is outside OpenClaw 2026.7.1's exact engine range by installing the SHA256-pinned Node 24.18.0 executable before the SHA512-pinned OpenClaw package.
- Replace Hermes' terminal-only configuration action with a searchable, responsive Settings UI generated from the exact installed runtime schema, verified live with Hermes Agent 0.20.0.
- Add typed, authenticated gateway routes for safe Hermes config and write-only credential management while preserving channel-owned secrets.
- Bridge Hermes 0.20's protected loopback dashboard endpoints through its supported `X-Hermes-Session-Token` contract using an owner-local mode-0600 credential file.
- Prevent invalid list/number drafts from saving stale values, preserve newer edits and user reverts across pending saves, merge save completions into freshly reloaded configuration, preserve unsaved text across navigation and reopening, and order credential refreshes so stale responses or redacted metadata cannot overwrite newer mutations.
- Keep agent installation visible in Terminal; this PR changes day-to-day Hermes configuration, not the visible installer contract.

Stack: 1/1 (based on `main`)

## Diagnosis

OpenClaw 2026.7.1 declares `node >=22.22.3 <23 || >=24.15.0 <25 || >=25.9.0`. Existing VPS upgrades retain `/opt/matrix/runtime/node`, so the observed 24.14.1 runtime is genuinely incompatible; lowering the compatibility gate would only defer the failure into npm. The installer now verifies and atomically replaces only the incompatible Node executable with the official pinned 24.18.0 build before continuing.

Hermes Agent 0.20.0 exposes configuration, defaults, schema, and environment endpoints, with its ephemeral dashboard session token required for sensitive loopback reads. Matrix now treats the installed runtime's schema as the UI source of truth instead of attempting to encode a moving command/field surface in the shell. The loopback-only dashboard wrapper derives Hermes' supported `HERMES_DASHBOARD_SESSION_TOKEN` from the random local credential secret, while the gateway reads that strict file with `O_NOFOLLOW` and sends `X-Hermes-Session-Token` only to `127.0.0.1`; the token never reaches the browser or logs.

The exact 0.20.0 runtime publishes 679 typed configuration fields. The gateway therefore accepts up to 1,024 fields—large enough for the verified version while retaining a finite resource bound—and filters sensitive paths before returning the schema to the browser.

## Tests

[...]

PR #1158, opened 2026-08-06
fix(agent): validate Hermes MoA preset readiness

xyhStruggler:
## Summary

- validate the active Hermes MoA preset instead of trusting the upstream provider-level `authenticated` flag
- require every enabled MoA reference provider and the aggregator provider to be authenticated before Matrix shows MoA as Ready
- fail closed to Action Required when the bounded MoA config response is missing, malformed, or incomplete
- add a visible Configure MoA action that opens `hermes moa configure` in the Matrix terminal

## Problem

Hermes reports the virtual `moa` provider as authenticated even when the selected preset's reference or aggregator providers have no credentials. Matrix therefore showed **Ready**, but the first real inference failed with `No LLM provider configured for task=moa_reference` or `task=moa_aggregator`.

The gateway now derives MoA readiness from the selected preset and the authenticated provider inventory. Ordinary non-MoA provider readiness is unchanged.

## Tests

- `flox activate -- bun run test -- tests/gateway/agent-config-hermes-source.test.ts tests/gateway/agent-runtime-services.test.ts tests/gateway/agent-settings-hermes-integration.test.ts tests/shell/agent-settings.test.tsx tests/shell/terminal-launch.test.ts` — 46 tests passed
- Hermes/Agent regression group — 121 tests passed across 15 files
- `flox activate -- bun run typecheck` — passed
- `flox activate -- bun run check:patterns` — passed with existing repository warnings and no violations
- focused ESLint for both changed shell files — passed
- `flox activate -- bun run build:shell:production` — passed
- `bun run test` — 10,421 passed, 20 skipped, and 24 unrelated macOS-local failures across existing Unix-socket, Linux-command, and shell timing tests; the MAT-259 lifecycle regression passes and the required Linux CI shards remain authoritative
- the first exact-head CI run exposed a stale lifecycle fixture that counted only two Hermes snapshot endpoints; commit `236c04aa` updates it to cover `/api/status`, `/api/model/options`, and `/api/config`

### Source of truth

- Hermes remains the source of truth for the selected model and MoA preset configuration.
- Matrix only derives a bounded readiness result from Hermes `/api/config` and `/api/model/options`; it does not copy credentials or configuration.

### Lock/transaction scope

- No database writes, locks, or transactions are introduced.
- The three loopback Hermes reads are independent, bounded, authenticated, and executed concurrently.

### Acceptable orphan states

- There are no new persisted objects or orphan states.
- If any MoA readiness input is unavailable or invalid, the UI fails closed to Action Required and keeps the saved provider/model selection intact.

### Auth source of truth

- The existing Matrix-to-Hermes dashboard token remains the authentication boundary.
- Provider secrets remain write-only and owned by Hermes; Matrix receives only provider authentication state.

### Deferred scope

- This does not change Hermes' upstream virtual-provider contract or mutate MoA presets.
- Configuring credentials remains an explicit visible Hermes terminal flow.

## Tracking

- [MAT-259](https://linear.app/matrix-os/issue/MAT-259/fix-false-ready-state-for-incomplete-hermes-moa-presets)
- Related: #1154, #1156

### DIFF
--- a/packages/gateway/src/agent-config/hermes-source.ts
+++ b/packages/gateway/src/agent-config/hermes-source.ts
@@ -17,6 +17,8 @@ export type HermesJsonReader = (
 const MAX_HERMES_PROVIDERS = 31;
 const MAX_HERMES_MODELS = 253;
 const MAX_MODELS_PER_PROVIDER = 128;
+const MAX_MOA_PRESETS = 64;
+const MAX_MOA_REFERENCES = 32;
 
 const ProviderIdSchema = z.string()
   .trim()
@@ -45,6 +47,21 @@ const HermesProviderSchema = z.object({
   auth_type: z.string().max(64).optional(),
   is_user_defined: z.boolean().optional(),
   models: z.array(z.unknown()).max(512).optional(),
+}).passthrough();
+
+const HermesConfigSchema = z.object({
+  moa: z.unknown().optional(),
+}).passthrough();
+
+const MoASlotSchema = z.object({
+  provider: ProviderIdSchema,
+  model: ModelIdSchema,
+  enabled: z.boolean().optional(),
+}).passthrough();
+
+const MoAPresetSchema = z.object({
+  reference_models: z.array(z.unknown()).max(MAX_MOA_REFERENCES),
+  aggregator: z.unknown(),
 }).passthrough();
 
 function authKindForProvider(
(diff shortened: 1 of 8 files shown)


xyhStruggler:
## Visual validation

Rendered locally from exact head `8ce0e0ef0ccb3c456f5d443b3752d5450e01c510` using the same bounded `AgentSettingsView` contract fixture covered by the shell regression test. This verifies the current component layout before the Preview VPS gate is enabled.

MoA Action Required with Configure MoA

Verified visually:
- Mixture of Agents is selected and shows **Action Required**.
- The incomplete OpenRouter dependency also shows **Action Required**.
- The selected model remains `default`.
- The visible **Configure MoA** action is present without exposing credentials or raw provider errors.

Preview VPS verification will follow after the review gate enables `ready-for-ci`.

@greptileai review

Please re-review the unchanged exact head above with this visual evidence included.

xyhStruggler:
## Exact-head validation update

The first Linux CI run found one stale lifecycle fixture: a Hermes snapshot now reads three bounded endpoints (`/api/status`, `/api/model/options`, and `/api/config`), while the existing away-and-back invalidation test still expected two.

Commit `236c04aa04ea004ebc5650f908df3052e55524ed` updates that regression coverage without changing production behavior.

Validation at this exact head:
- focused MAT-259 group: 46/46 passed
- typecheck: passed
- pattern scan: no violations
- full local suite: 10,421 passed, 20 skipped, 24 known macOS-specific failures unrelated to Hermes/MoA

The prior Preview VPS attempt failed in platform provisioning before the bundle was deployed. The `preview-vps` workflow is rebuilding and retrying against this exact head.

@greptileai review

Please re-review exact head `236c04aa04ea004ebc5650f908df3052e55524ed`.

xyhStruggler:
Closing this Matrix-side workaround after confirming the misleading MoA readiness originates in Hermes' upstream virtual-provider contract. The public documentation now records the current limitation and the required provider setup. We should address readiness at the Hermes source of truth rather than maintain a duplicate preset-validation layer in Matrix.

PR #1163, opened 2026-08-06
fix(desktop): reconcile project creation and folder reuse

xyhStruggler:
## Summary

- attach a stable, payload-scoped idempotency key to Desktop project and custom-folder creation
- replay timed-out creates beyond the gateway's bounded five-minute clone window
- reconcile overlapping folder requests through owner-scoped, bounded receipts
- publish folder project configs atomically without overwriting a concurrent winner
- open already-connected workspaces instead of showing a false create error
- distinguish Matrix-managed project registry directories from selectable repo workspaces
- recover stale slug conflicts after refreshing the authoritative project list

Closes MAT-263.

## UX behavior

- Selecting an already-connected workspace such as `apps/matrix-os` opens its existing project.
- Selecting `projects/<slug>` explains that it contains Matrix project data and offers to open the owning project.
- Matrix-managed registry metadata and worktree folders cannot be connected as user workspaces.
- A managed `projects/<slug>/repo` checkout remains selectable and resolves to its owning project by path.
- Reusing a slug for a genuinely different folder still asks the user to choose another name.

## Tests

- `bunx vitest run tests/desktop/create-project-dialog.test.tsx tests/desktop/files-browser-views.test.tsx tests/desktop/add-project-dialog.test.tsx` (57 passed)
- `bun run typecheck`
- `bun run build:desktop`
- `bun run check:patterns` (0 violations)
- live Electron validation against `pr-1163` for both `apps/matrix-os` and `projects/matrix-os`

Local full-suite note: `bun run test` completed with 10,354 passed and 24 failures in unchanged suites. The failures are outside this diff and include missing Vitest resolution in UI test setup, Codex control-socket and interactive-shell timeouts, a date-sensitive Todo fixture, and environment-specific host-tool tests. The changed Desktop suites passed in the same run; CI on the new PR head is authoritative.

### Source of truth

- `projects/<slug>/config.json` remains the canonical project record.
- Folder request receipts are bounded idempotency metadata only; a replay is accepted only when the receipt fingerprint, owner scope, and on-disk target all agree.
- The refreshed Desktop project list is authoritative when reconciling an existing workspace path.

### Lock/transaction scope

- The existing per-slug process lock still serializes local creates.
- Folder config publication additionally uses an atomic exclusive hard-link, so separate gateway processes cannot overwrite the winning config.
- Custom-folder creation uses exclusive `mkdir`; only an `EEXIST` from the same request fingerprint enters bounded receipt reconciliation. No network call is held inside these critical sections.

### Acceptable orphan states

- If custom-folder creation succeeds and later binding fails, the empty user folder may remain and can be connected through Existing folder.
- If project metadata directory creation succeeds but exclusive config publication fails, an empty metadata directory may remain; no existing config is overwritten.
- Receipts expire after 24 hours and are capped at 256 entries with symlink-safe cleanup.

### Auth source of truth

- The workspace route's request principal remains authoritative.
- The principal-derived owner scope is included in the folder request fingerprint and receipt; invalid or mismatched requests fail closed.

### Deferred scope

- This PR does not redesign project creation as an asynchronous job API.
- It does not migrate or delete previously duplicated user folders automatically.
- No public documentation change is proposed; this aligns Desktop recovery with the existing project registry contract.

### DIFF
--- a/desktop/src/renderer/src/features/board/AddProjectStepFields.tsx
+++ b/desktop/src/renderer/src/features/board/AddProjectStepFields.tsx
@@ -2,7 +2,8 @@
 // orchestration live in CreateProjectDialog; these components only render
 // fields and forward edits.
 import { Button } from "../../design/primitives";
-import ComputerFileBrowser from "../files/ComputerFileBrowser";
+import type { Project } from "../../stores/board";
+import ComputerFileBrowser, { type FolderPickerChoice } from "../files/ComputerFileBrowser";
 
 export const FIELD_STYLE: React.CSSProperties = {
   background: "var(--bg-raised)",
@@ -103,15 +104,45 @@ export function ExistingFolderStepFields({
   name,
   onNameChange,
   folderPath,
+  projects,
   onChooseFolder,
+  onOpenProject,
   submitting,
 }: {
   name: string;
   onNameChange: (value: string) => void;
   folderPath: string;
+  projects: Project[];
   onChooseFolder: (path: string) => void;
+  onOpenProject: (slug: string) => void;
   submitting: boolean;
 }) {
+  const projectForRegistryPath = (path: string): Project | undefined => {
+    const segments = path.replace(/^\.\/+/, "").replace(/\/+$/, "").split("/");
+    if (segments.length !== 2 || segments[0] !== "projects") return undefined;
+    return projects.find((project) => project.slug === segments[1]);
+  };
+  const resolveFolderChoice = (path: string): FolderPickerChoice => {
+    const normalized = path.replace(/^\.\/+/, "").replace(/\/+$/, "");
+    const segments = normalized.split("/");
+    if (segments[0] !== "projects") return { kind: "choose" };
+    // projects/<slug> is Matrix-owned registry state. Only the repo checkout
+    // beneath it (and its descendants) is a user workspace.
+    if (segments.length >= 3 && segments[2] === "repo") return { kind: "choose" };
+    const project = projectForRegistryPath(normalized);
+    if (project) {
+      return {
+        kind: "alternate",
+        label: `Open ${project.name}`,
+        message: "This folder contains Matrix project data, not a workspace.",
+      };
+    }
+    return {
+      kind: "blocked",
+      message: "This folder is managed by Matrix and can't be used as a workspace.",
+    };
+  };
+
   return (
     <>
       <label className="flex flex-col gap-1">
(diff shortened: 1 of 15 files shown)


xyhStruggler:
Fixed the 4/5 receipt-expiration finding in `68bedf57`.

Folder request receipts are now validated against their persisted `createdAt` on every read. Malformed or older-than-24-hour receipts are removed before replay/conflict checks, so an expired key no longer replays indefinitely or blocks a later payload. Added a regression test that rewrites a receipt past its TTL and verifies the same request ID can create a new folder payload.

Fresh verification: 210 focused tests, full typecheck, desktop production build, and pattern scan all pass.

@greptileai review

xyhStruggler:
Fixed the overlapping expiration-cleanup race in `0b1943b7`.

Folder receipt renewal now runs under the shared request-key lock, covering the full read/remove/create/publish sequence. This prevents a delayed cleanup from unlinking the replacement receipt produced by an overlapping retry. The new regression test deliberately holds concurrent receipt removals open, verifies cleanup is serialized, confirms both requests succeed, and confirms a later retry replays the persisted result.

Fresh verification: 211 focused tests, full typecheck, desktop production build, and pattern scan all pass.

@greptileai review

xyhStruggler:
Added exact regression coverage for the live preview failure in `2361980f`. The test pre-creates an unmanaged `projects/matrix-os` registry directory, creates `apps/matrix-os` through the mkdir route, and verifies the current manager atomically publishes project config and connects the custom folder instead of returning the legacy `slug_conflict`.

Focused desktop/gateway verification: 212 tests passed.

@greptileai review

xyhStruggler:
Preview validation is currently blocked by VPS provisioning infrastructure, not by the MAT-263 bundle. The preview workflow successfully built and registered `v2026.08.06-pr1163-2361980`, but `pr-1163` remained `provisioning` for the full 10-minute wait on attempts 1 and 3; attempt 2 eventually transitioned to `failed`. The workflow was retried through the supported idempotent failed/provisioning path three times and never reached bundle deployment. Run: https://github.com/HamedMP/matrix-os/actions/runs/31112799766

The existing `pr-1156` runtime was intentionally left untouched because it runs a different preview bundle. Local regression coverage for the exact occupied `projects/matrix-os` case passes, and all PR CI plus Greptile 5/5 remain green.

xyhStruggler:
Post-outage retry results for head `88550652`:

- CI rerun: passed.
- Docker Tests rerun: passed.
- Preview bundle build: passed (`v2026.08.08-pr1163-8855065`).
- Preview provisioning: blocked before deploy. Fleet reports `pr-1163` as `failed`, and `POST /vps/preview/provision` now returns HTTP 409 immediately.

The workflow performed the requested single retry and did not touch `pr-1156`. Based on the platform retry path, this is consistent with an exhausted failed-machine retry budget or another preview allocation conflict. The current preview workflow has no manual teardown dispatch, so `pr-1163` needs operator cleanup/deletion before it can be provisioned fresh. Run: https://github.com/HamedMP/matrix-os/actions/runs/31263033891

xyhStruggler:
@greptileai

xyhStruggler:
Ready for review.

The latest update reconciles existing-folder selection across Desktop project locations:

- already-connected workspaces open the existing project
- Matrix-managed `projects/<slug>` registry directories are no longer connectable as workspaces
- managed `repo` workspaces remain selectable and resolve by path
- stale project-list conflicts recover after refresh

Validation:

- 57 targeted Desktop tests passed
- repository typecheck passed
- Desktop production build passed
- pattern scan passed with 0 violations
- verified live in Electron against `pr-1163`

The new head is `78f8cd32`; CI and Preview VPS validation are running.

Nima-Naderi (approved):

PR #1166, opened 2026-08-06
fix(release): make fleet deployment opt-in

Nima-Naderi:
## Summary

- reapply PR #1160 under the intended GitHub account
- restore normal main pushes that publish and promote dev
- keep deployment explicitly opt-in through trusted manual dispatch only
- remove the temporary revert interlock

## Tests

- red contract test against the temporary interlock
- workflow and host-bundle tests: 78 passed
- replacement tree matches the original #1160 final tree exactly

## Review/Monitoring

- second PR in the stack, based on the combined revert PR
- merge only after the combined revert
- verify the post-merge host-bundle workflow skips fleet deployment

## Invariants

- source of truth: platform Postgres remains authoritative for release and channel pointers
- lock/transaction scope: no database writes are introduced by this change
- acceptable orphan states: publication may succeed while the optional golden-snapshot enqueue fails; fleet deployment remains separately authorized
- auth source of truth: PLATFORM_SECRET remains the deploy API credential and is never exposed
- deferred scope: persistent channel reporting and passive downgrade protection return in the next stacked PR

### DIFF
--- a/.github/workflows/host-bundle-release.yml
+++ b/.github/workflows/host-bundle-release.yml
@@ -1,8 +1,7 @@
 # Host Bundle Release
 #
 # Builds, publishes, and registers the Matrix OS host bundle.
-# TEMPORARY REVERT INTERLOCK: push releases are disabled until the deployment
-# containment change is reapplied in the next stacked PR.
+# Main publishes the dev channel. Tags publish immutable release versions.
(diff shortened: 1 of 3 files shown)


Nima-Naderi:
@greptileai

PR #1169, opened 2026-08-07
fix(observability): reduce PostHog log and Cloudflare 4xx noise

HamedMP:
## Root cause

- `matrix-symphony.service` renders an ANSI terminal dashboard every refresh. Alloy forwarded every redraw to both Loki and PostHog, producing roughly 21k low-value log records/minute during the observed burst.
- Disconnected Linear trackers poll `/internal/containers/<handle>/integrations/call` every 5 seconds. The expected setup-required 404 therefore creates about 17,280 Cloudflare 4xx requests per host/day, matching the dashboard's ~17.1k per-path counts.

## Fix

- Disable the Symphony status dashboard when no real terminal is attached (systemd/journald).
- Keep the Symphony journal source in Loki but exclude it from PostHog OTLP log shipping.
- Back off setup-required tracker polling to 5 minutes while preserving the configured interval for connected trackers.

This cuts disconnected polling by 60x (17,280 to 288 requests/host/day) and removes the dominant terminal-redraw stream from PostHog.

## Verification

- `pnpm exec vitest run tests/observability/posthog-logship.test.ts tests/deploy/customer-vps/symphony-systemd.test.ts`
- 2 files passed, 16 tests passed
- `git diff --check`

### DIFF
--- a/distro/customer-vps/host-bin/matrix-install-logship
+++ b/distro/customer-vps/host-bin/matrix-install-logship
@@ -168,7 +168,9 @@ loki.source.journal "matrix_symphony" {
   matches       = "_SYSTEMD_UNIT=matrix-symphony.service"
   relabel_rules = loki.relabel.journal.rules
   labels        = { source = "journal", handle = "${HANDLE}", env = "${MATRIX_ENV}" }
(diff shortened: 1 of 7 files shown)


HamedMP:
@greptileai review

PR #1205, opened 2026-08-12
fix(terminal): scale session hover cards with canvas zoom

HamedMP:
## Summary
- scale the terminal session hover card with the current terminal renderer's effective Canvas zoom even though Radix renders it through a body portal
- use the scaled dimensions when deciding whether the card fits to the right of its session row
- keep Desktop, mobile, and fullscreen terminal cards at scale 1 even when persistent Canvas state retains another zoom
- add regression coverage for 50% Canvas zoom and stale global zoom

## Root cause
The terminal window is inside the transformed Canvas tree, but the Radix hover card is portaled to `document.body`. It therefore escaped the Canvas transform and stayed at its unzoomed CSS-pixel size while the terminal window changed size. The first fix read persistent global Canvas zoom directly, which did not represent the effective renderer scale in Desktop, mobile, or fullscreen Canvas terminals.

## Tests
- [x] `pnpm exec vitest run tests/shell/terminal-session-hover-card.test.tsx` (5 tests)
- [x] `pnpm exec vitest run tests/shell/terminal-session-hover-card.test.tsx tests/shell/terminal-app-component.test.tsx --maxWorkers=1 --no-file-parallelism` (101 tests)
- [x] `bun run typecheck`
- [x] `bun run check:patterns` (0 violations)
- [x] `git diff --check`
- [x] label-gated GitHub CI: type check, pattern scan, React Doctor, shell production build, sync client package, all four unit-test shards, E2E, aggregate CI, Docker image build, and Docker smoke
- [ ] `bun run test` was attempted locally and stopped after unrelated baseline/environment failures in default-app suites (Node `localStorage` unavailable) and unrelated timeout cases; the changed terminal suites passed within the run

## Review/Monitoring
- addressed Greptile's 4/5 P1 and Codex's matching P2 by threading `TerminalApp.canvasZoom` through the sidebar/session-card path
- latest trusted Greptile review approved commit `18fd4f8ac13239f29e3993688492a733c03181fc` at an exact 5/5; both prior threads are resolved and outdated
- re-added `ready-for-ci` only after the 5/5 review, then monitored the triggered CI and Docker workflows through successful completion

## Invariants
- **Source of truth:** `TerminalApp.canvasZoom` is the terminal renderer's effective scale; `CanvasWindow` derives it from Canvas zoom and forces 1 for fullscreen, while Desktop and mobile use the default 1.
- **Lock/transaction scope:** Not applicable; UI-only change with no persistence.
- **Acceptable orphan states:** None; the hover card is still controlled by the existing Radix root and session state.
- **Auth source of truth:** Unchanged; no auth or network behavior is touched.
- **Deferred scope:** Other global shell overlays are unchanged; this PR fixes the reproduced terminal session hover-card path.

### DIFF
--- a/shell/src/components/terminal/TerminalApp.tsx
+++ b/shell/src/components/terminal/TerminalApp.tsx
@@ -936,7 +936,7 @@ export function TerminalApp({ initialCommand, initialLabel, initialClaudeMode = 
           className={mobile ? "relative flex flex-1 min-h-0 flex-col" : "relative flex flex-1 min-h-0"}
           style={{ background: "var(--terminal-app-body-bg)" }}
         >
-          <LocalTerminalSidebar />
+          <LocalTerminalSidebar canvasZoom={canvasZoom} />
           {activeTab ? (
             <div
               data-testid="terminal-content-surface"
--- a/shell/src/components/terminal/TerminalSessionHoverCard.tsx
+++ b/shell/src/components/terminal/TerminalSessionHoverCard.tsx
@@ -3,7 +3,6 @@
 import { useEffect, useState, type CSSProperties, type ReactElement, type RefObject } from "react";
 import { HoverCard as HoverCardPrimitive } from "radix-ui";
 import { SHELL_Z_INDEX } from "@/lib/shell-layering";
-import { useCanvasTransform } from "@/hooks/useCanvasTransform";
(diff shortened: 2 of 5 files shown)


Nima-Naderi:
@greptileai

Nima-Naderi (approved):
LGTM

PR #1261, opened 2026-08-19
fix(shell): restore reliable terminal input focus

Nima-Naderi:
## Summary

- recover Canvas pointer ownership after screen-recording and tab-focus transitions
- request xterm DOM focus for new, restored, switched, repeated, and replacement Terminal tabs
- preserve the existing non-focus-stealing background-session behavior
- add focused regression coverage for both failure modes

Fixes MAT-442.

## Tests

- PASS: pinned Node 24/Flox focused suite — 4 files, 120 tests
- PASS: full repository typecheck
- PASS: pattern scan — 0 violations, 5 existing warnings
- PASS: git diff check
- PASS: CI-discovered Terminal focus ref boundary now ignores non-focusable unknown values safely; the original Next build type error is absent from local shell TypeScript output
- BASELINE BLOCKED: broad unit suite — 43 files failed / 911 passed / 3 skipped; 304 tests failed / 10,645 passed / 20 skipped, primarily localStorage environment failures plus unrelated CLI/platform timeouts
- BASELINE BLOCKED: existing TerminalPane scroll suite — 27/29 failures because its xterm mock does not intercept the dynamic import after a frozen install and real xterm requires window.matchMedia; the unchanged suite fails before changed behavior
- BASELINE BLOCKED: shell lint cannot resolve next/dist/compiled/babel/eslint-parser
- BASELINE BLOCKED: local production build and runtime fail before changed modules compile in the local Mermaid/Langium dependency tree, including missing vscode-languageserver-types and internal export mismatches

## Review and monitoring

- Greptile review required at 5/5 for the latest head before ready-for-ci
- ready-for-ci was removed after the first CI run exposed the stricter Next build type boundary
- ready-for-ci will be restored only after the new head reaches Greptile 5/5
- fresh exact-head manual recording remains pending until the unrelated local shell dependency baseline compiles

## Invariants

- Source of truth: canonical shell sessions remain in /api/terminal/sessions; this patch adds only ephemeral UI focus requests.
- Lock and transaction scope: no database, lock, persistence, or transaction behavior changes.
- Acceptable orphan states: none introduced; focus request counters are component-local and disappear with the Terminal surface.
- Auth source of truth: unchanged; existing gateway and shell session authentication remains authoritative.
- Deferred scope: PTY and WebSocket lifecycle, session architecture, hover-card layout, clipboard behavior, and the unrelated local dependency baseline are intentionally unchanged.

### DIFF
--- a/shell/src/components/canvas/CanvasTransform.tsx
+++ b/shell/src/components/canvas/CanvasTransform.tsx
@@ -37,7 +37,9 @@ export function CanvasTransform({
   const scrollTimeoutRef = useRef<ReturnType<typeof setTimeout> | null>(null);
   const lastPointer = useRef({ x: 0, y: 0 });
   const spaceDown = useRef(false);
+  const interactionOverlayActiveRef = useRef(false);
   const [grabCursor, setGrabCursor] = useState(false);
+  const [interactionOverlayActive, setInteractionOverlayActive] = useState(false);
 
   useEffect(() => {
     const el = containerRef.current;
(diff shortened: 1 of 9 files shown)


Nima-Naderi:
@greptileai

PR #1278, opened 2026-08-20
fix(platform): normalize preview provider image ids

Nima-Naderi:
## Summary
- normalize Postgres `BIGINT` provider image IDs before exact preview-snapshot comparison
- preserve fail-closed behavior for invalid or unsafe provider IDs
- cover the production-shaped string ID returned by node-postgres

## Why
Production returned `provider_image_id` as a string while the exact preview guard compared it strictly to a number. The identical provider image was rejected, so the create intent was never persisted and Hetzner was never called.

## Tests
- TDD red: new bigint regression failed before implementation
- `pnpm exec vitest run tests/platform/golden-snapshot-provisioning.test.ts --reporter=dot` (37/37)
- `bun run typecheck`
- `pnpm --filter @matrix-os/platform exec tsc --noEmit -p tsconfig.typecheck.json`
- `bun run check:patterns:diff` (0 violations)
- `git diff --check`

## Review / monitoring
- Disposable v988 reproduction logged `internalReason=create_intent_unavailable`
- Production DB showed `provider_image_id` as `"422202690"` and no create intent
- Customer golden-snapshot rollout remains disabled at 0%

## Invariants
- **Source of truth:** the persisted provider image ID bound to the exact test snapshot
- **Lock / transaction scope:** existing snapshot, lease, and create-intent transaction is unchanged
- **Acceptable orphan states:** unchanged; no provider call occurs until the create intent is durably linked
- **Auth source of truth:** existing operator-only preview provisioning authorization is unchanged
- **Deferred scope:** customer snapshot rollout remains disabled; this PR only repairs the operator exact-snapshot test path

### DIFF
--- a/packages/platform/src/golden-snapshot-preview-test.ts
+++ b/packages/platform/src/golden-snapshot-preview-test.ts
@@ -308,6 +308,14 @@ const CreateIntentInputSchema = z.object({
   now: IsoDateSchema,
 }).strict();
 
+const PreviewProviderImageIdSchema = z.coerce.number().int().positive()
+  .max(Number.MAX_SAFE_INTEGER);
+
+export function normalizePreviewTestProviderImageId(value: unknown): number | undefined {
+  const parsed = PreviewProviderImageIdSchema.safeParse(value);
+  return parsed.success ? parsed.data : undefined;
+}
+
 export async function createPreviewTestSnapshotCreateIntent(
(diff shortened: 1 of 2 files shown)


Nima-Naderi:
[codex] @greptileai review latest head c8a596526f55f58e78a1ff303ecfe39ce314d156

PR #1298, opened 2026-08-22
feat(integrations): register local MCP across agents

shubhamzanwar:
## Summary

- Ship `/opt/matrix/bin/matrix-integrations-mcp`, a local stdio MCP server with safe inventory, list, describe, connect, sync, call, and disconnect tools.
- Register the Matrix-owned MCP entry automatically for Codex, Claude Code, Hermes, and OpenClaw after installation and on gateway boot.
- Give OpenCode, Pi, and other terminal agents the same gateway contract through the auto-discovered Matrix skill and credential-isolated `matrix-integrations` command.
- Keep provider credentials behind the existing VPS gateway -> platform integration proxy and start agent helpers with an explicit environment allowlist.
- Document the updated architecture, comparison with the earlier state, security boundary, and deployed-PR test procedure.
- Leave GitHub workflows unchanged.

## Tests

- 27/27 new MCP protocol, command fallback, handler, and VPS registration tests pass.
- 310/312 targeted integration, host-bundle, cloud-init, OpenClaw, installer, and agent-launcher tests pass.
- The same two `customer-vps-cloud-init.test.ts` tests fail unchanged on the base workspace on macOS because they assume GNU `mv -T` and GNU `timeout` behavior.
- `pnpm run check:patterns:diff` (0 violations; existing join/header warnings reviewed).
- Kernel, integrations-mcp, and gateway TypeScript checks pass.
- No `.github/workflows` files are changed.

### Source of truth

- Platform Postgres remains the source of truth for user-scoped connected accounts; Pipedream remains the OAuth/provider-token custodian behind the platform boundary.
- Agent MCP configurations are projections of the Matrix-owned `matrix-integrations` entry. VPS bootstrap replaces only that named entry and retries on gateway restart.

### Lock/transaction scope

- This change adds no database write sequence or lock. Existing platform connection ownership, upsert, action validation, and disconnect transactions remain authoritative.
- MCP and command handlers perform one bounded gateway request per operation; provider network calls remain in the existing platform integration service.

### Acceptable orphan states

- If an agent CLI is absent, registration skips it. If registration fails after removing the previous Matrix-owned entry, that agent temporarily has no Matrix MCP entry; gateway boot and the installer retry registration without affecting user-owned MCP servers.
- A failed MCP process or gateway request creates no connection state. Existing OAuth sync reconciles an account authorized before a webhook or client failure.

### Auth source of truth

- The customer VPS gateway validates its Matrix bearer token; the platform internal route validates the VPS internal token and resolves the user/account owner server-side.
- Launchers use hard-coded loopback/runtime paths, read only the Matrix token and user ID from host configuration, and start from an empty environment. Provider, database, storage, and analytics credentials are not inherited.

### Deferred scope

- User-built app SDK permissions and per-app integration grants are intentionally separate from this updated agent architecture.


shubhamzanwar:
@greptileai review

shubhamzanwar:
@greptileai review

shubhamzanwar:
@greptileai review

PR #1304, opened 2026-08-24
fix(codex): verify 0.149.1 provider contracts

shubhamzanwar:
## Summary

- record `@openai/codex` 0.149.1 as the latest verified exec JSONL and app-server contract
- align the shared runtime version source, Docker image pin, and customer-VPS installers with 0.149.1
- update gateway contract assertions so 0.149.1 is explicitly accepted as verified

## Tests

- upstream 0.149.1 schema verification via `scripts/check-codex-exec-contract.mjs`
- focused Codex gateway, runtime, Docker, VPS installer, and contract tests: 160 passed
- `./scripts/review/check-patterns.sh --diff origin/main`: 0 violations (one pre-existing `Set` warning in `packages/contracts/src/index.ts`)
- cloud-init focused suite: 2 unrelated macOS failures (`mv -T` and GNU `timeout` are unavailable); its 41 other tests passed
- full `bun run typecheck && bun run check:patterns && bun run test` not run because `bun` is unavailable on the local PATH

### Source of truth

- `CODEX_VERIFIED_VERSION` is the runtime/install source of truth and must match both verified-contract ledgers.
- The scheduled workflow resolves the latest published package and independently verifies the ledgers against upstream artifacts.

### Lock/transaction scope

- This is a static configuration-only update; it creates no runtime lock, transaction, or network call.

### Acceptable orphan states

- No artifacts or persistent runtime state are created. An unverified release stays blocked by the exact-version gate and scheduled contract check.

### Auth source of truth

- No authentication behavior changes. The existing provider executable validation remains the runtime authority.

### Deferred scope

- This PR does not alter event parsing, app-server protocol handling, or the verification workflow. The upstream 0.149.1 schemas match the previously accepted artifacts.


HamedMP (approved):

Nima-Naderi (approved):

shubhamzanwar:
@greptileai review current head: 6e9ce13

PR #1345, opened 2026-08-27
feat(desktop): add native show desktop animation

shubhamzanwar:
## Summary

- Add native Electron Show Desktop behavior to the Desktop title-bar tab and empty desktop surface.
- Keep floating OS windows mounted while animating each to a stable random top, left, or right edge, leaving 12px visible.
- Exclude full-screen tab surfaces from Show Desktop and keep them visible.
- Restore each window to its original window/tab presentation on the next toggle.

## Tests

- `pnpm exec vitest run tests/desktop/desktop-surfaces-store.test.ts tests/desktop/native-desktop-shell.test.tsx` — 49 passed
- `pnpm exec tsc --noEmit -p desktop/tsconfig.json`
- `pnpm run check:patterns` — passed with existing baseline warnings
- `pnpm run typecheck` — could not start because `bun` is unavailable in this environment
- `pnpm run test` — package builds passed, but the full Vitest run produced no results after several minutes and was interrupted

### Source of truth
- Native renderer desktop surface store is the source of truth for the show-desktop transition and hidden surface IDs.
- No new persistence or second data store was introduced.

### Lock/transaction scope
- Not applicable; this is renderer-only state and animation.

### Acceptable orphan states
- None; OS windows remain mounted during and after the animation and restore from their original modes.

### Auth source of truth
- Unchanged; no auth paths were modified.

### Deferred scope
- Web shell and Canvas mode are intentionally unchanged; this PR is native Electron Desktop mode only.

Closes MAT-496

### DIFF
--- a/desktop/src/renderer/src/design/index.css
+++ b/desktop/src/renderer/src/design/index.css
@@ -6,6 +6,16 @@
 @import "@fontsource/jetbrains-mono/400.css";
 @import "@fontsource/jetbrains-mono/500.css";
 @import "./tokens.css";
+
+@keyframes native-desktop-window-hide {
+  from { opacity: 1; transform: translate3d(0, 0, 0); }
+  to { opacity: 1; transform: translate3d(var(--desktop-exit-x), var(--desktop-exit-y), 0); }
+}
+
+@keyframes native-desktop-window-show {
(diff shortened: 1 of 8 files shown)


Nima-Naderi (approved):
🚀

shubhamzanwar:
@greptileai review again

PR #1350, opened 2026-08-27
fix(vps): self-heal missing host Node runtime

shubhamzanwar:
## Summary

- add a serialized repair helper that restores the checksum-verified bundled Node runtime atomically while preserving non-bundled global developer tools
- keep the privileged repair path metadata-only: it never executes the installed or staged Node binary, and it restores `bin/node` as root-owned/non-writable behind protected parent rename boundaries
- run the helper before gateway startup and integrations registration, which lets an older updater repair the runtime from its still-staged bundle before the gateway health check
- extract unprivileged runtime process checks and exact-version repair scheduling into `matrix-host-runtime-state`, keeping this filesystem lifecycle out of the 1,000+ line sync agent

## Tests

- shell syntax checks for all changed host scripts
- 117 focused Vitest tests covering host bundles, self-host installation, integration registration, and systemd wiring
- `pnpm run check:patterns` (0 violations; existing repository warnings only)
- `git diff --check`

## Release

After merge, validate the automatically published `dev` host bundle on one affected/disposable VPS. Then manually dispatch **Host Bundle Release** from `main` with `channel=stable`, `severity=normal`, a one-line changelog, and `deploy_after_publish=true` only after the exact dev bundle has passed live verification.

### Source of truth

- The immutable, checksum-verified host bundle remains the source of truth for the base Node runtime.
- `/opt/matrix/runtime/node` remains the installed runtime prefix. Global packages and user-installable CLI entries stay matrix-writable, while the prefix and bundle-owned `bin/node` remain root-controlled.

### Lock/transaction scope

- A root-owned `flock` serializes staging copies and atomic activation across concurrent systemd pre-start calls.
- Bundle download, checksum verification, and unprivileged process-level readiness checks remain outside this lock in the sync agent/state helper.

### Acceptable orphan states

- Interrupted copies may leave PID-scoped `.node.incoming.*` or `.node.previous.*` directories; the active runtime is switched only by same-filesystem rename, and the helper restores the previous directory if activation or metadata verification fails.
- A repaired runtime may remain installed if a later app update step rolls back. This is intentional because it restores a required host prerequisite without changing owner data.

### Auth source of truth

- No auth boundary changes. Platform release metadata and signed bundle URLs remain the update authorization source.
- The repair helper accepts runtime payloads only from the local update staging tree and must run as root. It does not execute binaries from that staging tree.

### Deferred scope

- This does not change release-channel policy or automatically authorize a fleet rollout.
- This does not replace or upgrade an already healthy Node runtime; versioned base-runtime upgrades remain separate work.


shubhamzanwar:
@greptileai review

shubhamzanwar:
@greptileai review latest changes again

PR #1352, opened 2026-08-27
fix(codex): verify 0.150.1 provider contracts

shubhamzanwar:
## Summary

- record `@openai/codex` 0.150.1 as the latest verified exec JSONL and app-server contract
- align the shared runtime version source, production and development Docker image pins, and customer-VPS installers with 0.150.1
- add gateway and development-container contract assertions so every managed install path accepts the verified release

## Tests

- upstream 0.150.1 schema verification via `scripts/check-codex-exec-contract.mjs`
- focused Codex gateway, runtime, Docker, VPS installer, and contract tests: 121 passed
- `./scripts/review/check-patterns.sh --diff origin/main`: clean
- full `bun run typecheck && bun run check:patterns && bun run test` not run because `bun` is unavailable on the local PATH

### Source of truth

- `CODEX_VERIFIED_VERSION` is the runtime/install source of truth and matches both verified-contract ledgers and every managed Docker/VPS installer pin.
- The scheduled workflow resolves the latest published package and independently verifies both ledgers against upstream artifacts.

### Lock/transaction scope

- This is a static configuration-only update; it creates no runtime lock, transaction, or network call.

### Acceptable orphan states

- No artifacts or persistent runtime state are created. An unverified release remains blocked by the exact-version gate and scheduled contract check.

### Auth source of truth

- No authentication behavior changes. The existing provider executable validation remains the runtime authority.

### Deferred scope

- This PR does not alter event parsing, app-server protocol handling, or the verification workflow. The upstream 0.150.1 schemas match the previously accepted 0.150.0 artifacts.


Nima-Naderi (approved):

shubhamzanwar:
@greptileai review again

PR #1354, opened 2026-08-27
fix(ci): retry change-detection checkout fetches

### LINKED ISSUE #1254: ci: retry transient checkout failures in change detection
## Problem

Main CI run https://github.com/HamedMP/matrix-os/actions/runs/32141074987 was cancelled before tests executed because the Detect CI-relevant changes job stalled during actions/checkout fetch. The job reached its two-minute timeout, which skipped all required jobs and caused CI Results to fail.

The downstream Host Bundle Release then failed its same-SHA gate:
https://github.com/HamedMP/matrix-os/actions/runs/32141074960

A transient runner or network stall should not fail the entire main release pipeline after one fetch attempt.

## Proposed fix

Add a bounded retry for the change-detection checkout or fetch operation.

- Retry only transient checkout or fetch failures.
- Use a small fixed maximum, such as two or three total attempts.
- Apply bounded backoff between attempts.
- Preserve a hard overall job timeout so a genuinely stuck fetch terminates.
- Log the attempt number and final failure clearly.
- Do not allow retries to trigger duplicate test or release runs.

## Acceptance criteria

- A simulated transient first-attempt fetch failure is retried and can recover.
- Retry count and total execution time are strictly bounded.
- Permanent checkout failures still fail the change-detection job.
- Existing source-change and docs-only detection behavior remains unchanged.
- Workflow contract coverage protects the retry limit and timeout.
- Same-SHA Host Bundle gating proceeds normally after a recovered CI run.

Nima-Naderi:
## Summary

- replace the one-shot change-detection checkout with three shallow fetch attempts
- cap each fetch at 30 seconds, apply 5s/10s bounded backoff, and preserve a three-minute job timeout
- clear only known Git lock files after failed attempts and log every attempt/final failure
- execute the workflow checkout script in contract tests to prove transient recovery and bounded permanent failure

Closes #1254

## Tests

- `pnpm exec vitest run tests/platform/ci-workflows.test.ts` (36/36 passed)
- `bun run typecheck` (passed)
- `bun run check:patterns` (0 violations; existing warnings only)
- `bun run test` (broad run stopped after five unrelated timeouts in `tests/platform/proxy-routing.test.ts`; changed-area suite was rerun and passed afterward)

## Review/Monitoring

- Greptile current-head review: 5/5 on `c2f47df6808ca72088e70677bd064e8a3f31900a`
- Claude review: passed
- label-gated CI: changed checkout job, typecheck, pattern scan, shell production build, sync client, and three unit shards passed
- residual CI failure: `tests/gateway/session-registry.test.ts` expects `/bin/zsh` when the runner's default shell is `/bin/bash`; reproduced locally and unrelated to this PR's workflow/test files
- `ready-for-ci` removed while the aggregate CI gate remains red

## Invariants

- **Source of truth:** `.github/workflows/ci.yml` remains the sole CI change-detection and downstream fan-out definition.
- **Lock/transaction scope:** no database or application-state writes; retry cleanup is limited to known Git lock files in the ephemeral runner checkout.
- **Acceptable orphan states:** a failed attempt may leave shallow-fetch locks, which are removed before retry; downstream test and release jobs cannot start until change detection succeeds.
- **Auth source of truth:** `${{ github.token }}` is passed only to the fetch command through an HTTP authorization header and is not written to repository config.
- **Deferred scope:** other workflow checkout steps keep their existing behavior; this PR targets the main CI change-detection failure from #1254.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -41,13 +41,13 @@ jobs:
             if timeout --foreground --kill-after=5s "${CHECKOUT_TIMEOUT_SECONDS}s" \
               git -c http.extraheader="AUTHORIZATION: basic $auth_header" \
                 fetch --no-tags --prune --no-recurse-submodules --depth=1 \
-                origin "+${GITHUB_REF}:refs/remotes/origin/ci" \
+                origin "$GITHUB_SHA" \
               && git checkout --force --detach "$GITHUB_SHA"; then
               echo "Checkout fetch succeeded on attempt $attempt/$CHECKOUT_MAX_ATTEMPTS"
               exit 0
             fi
 
-            rm -f .git/index.lock .git/shallow.lock .git/refs/remotes/origin/ci.lock
+            rm -f .git/index.lock .git/shallow.lock
             if [ "$attempt" -eq "$CHECKOUT_MAX_ATTEMPTS" ]; then
               echo "::error::Checkout fetch failed after $CHECKOUT_MAX_ATTEMPTS attempts"
               exit 1
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -17,14 +17,46 @@ jobs:
   changes:
     name: Detect CI-relevant changes
     runs-on: ubuntu-latest
-    timeout-minutes: 2
+    timeout-minutes: 3
     outputs:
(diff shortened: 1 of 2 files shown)


Nima-Naderi:
@greptileai

PR #1401, opened 2026-08-29
fix(security): isolate customer R2 backups

HamedMP:
## Summary

- remove standing R2/S3 provider credentials from customer VPS config, cloud-init, systemd, and matrixctl
- broker backup/restore through authenticated platform routes that derive the tenant prefix server-side and allowlist exact system keys
- preserve large database dumps with streamed multipart upload, atomic downloads, and missing-object exit code 44
- migrate legacy VPSes by removing only the exact regular non-symlink r2.env after release health and broker readiness, with startup and five-minute retries
- update operator security guidance and threat-model status

## Security invariant

A customer VPS never receives a standing provider storage credential and cannot request a presigned capability outside the active VPS owner's server-derived prefix and exact backup-key allowlist.

## Required invariants

- **Source of truth:** platform Postgres maps the authenticated handle to the active Clerk owner; platform R2 configuration owns the bucket and prefix root. Customer-supplied prefixes are never authorization inputs.
- **Lock/transaction scope:** no database mutations are added. Storage operations are capability-scoped; latest-pointer publication remains ordered after snapshot completion.
- **Acceptable orphan states:** an interrupted multipart upload may remain only when best-effort abort also fails; bucket multipart lifecycle cleanup remains the backstop. A local dump is retained when broker upload fails.
- **Auth source of truth:** constant-time HMAC(handle, PLATFORM_SECRET) bearer verification plus active VPS ownership lookup, with legacy container lookup only as fallback.
- **Deferred scope:** fleet rollout and provider-token rotation are operational and must follow merge. The public matrix-os-site documentation update remains a separate private-repository deliverable. The existing 1,329-line sync agent should have update-phase and migration orchestration extracted in a follow-up refactor; this urgent patch keeps the new migration isolated in named functions and tests.

## Rollout gate

Do not mark the incident closed on merge alone:

1. deploy the platform broker before or with the host bundle
2. verify backup upload, latest-pointer read, and restore probe on reviewed VPSes
3. verify /opt/matrix/env/r2.env is absent fleet-wide
4. rotate the platform R2 credential and revoke the formerly shared token

Historical cloud-init or copied credentials remain usable until step 4.

## Verification

- 258 focused storage, gateway, provisioning, cloud-init, host-bundle, and R2-client tests passed
- platform TypeScript build passed
- shell and Node syntax checks passed
- mandatory pattern scanner: 0 violations
- fresh Codex Security bypass review completed; its first-OTA migration and production multipart findings were reproduced with red tests and fixed


HamedMP:
@greptile review

PR #1480, opened 2026-08-31
fix(shell): refresh stale clients and open support chat

HamedMP:
## Summary
- retire the stale executable-asset service-worker caches and coordinate one safe reload when an updated worker takes control
- open the actual PostHog Conversations panel from the web navbar and preserve early Electron support clicks while the widget initializes
- make the default shell HTTP fetch fallback visibly timeout-bound so the same-SHA Pattern Scan no longer blocks Host Bundle Release

## Root cause
- the v2 workers cached JavaScript/CSS cache-first and the web registration left updated workers waiting, allowing mixed release generations until users cleared Cache Storage
- PostHog Conversations `show()` restores/displays the widget but does not guarantee the panel is open; web never clicked the provider launcher
- Electron rejected support opens before PostHog identity initialization completed
- the HTTP wrapper passed a timeout signal at request time, but the fallback `fetch()` itself did not expose one within the release scanner window

## Verification
- 30 focused Vitest tests
- shell, Desktop renderer/main, and platform TypeScript checks
- Electron production build
- Next.js production shell build
- `./scripts/review/check-patterns.sh` (0 violations)
- `git diff --check`

## Invariants
- **Source of truth:** browser/CDN HTTP caching owns executable Next.js assets; the service worker owns only versioned shell HTML fallback plus safe media/font assets. PostHog Conversations remains the support UI provider.
- **Lock/transaction scope:** none; this change has no database writes or shared persistence mutation.
- **Acceptable orphan states:** if service-worker activation fails, the current controller remains and registration retries on a later load. If support cannot initialize/open within its bounded timeout, it returns false and Electron releases its renderer overlay.
- **Auth source of truth:** unchanged; Clerk/runtime connection identity and the existing build-time PostHog project token remain authoritative.
- **Deferred scope:** no production bundle publish, VPS rollout, analytics-dashboard changes, or unrelated service-worker redesign is included.

### DIFF
--- a/desktop/src/renderer/src/features/support/DesktopSupportWidget.tsx
+++ b/desktop/src/renderer/src/features/support/DesktopSupportWidget.tsx
@@ -107,6 +107,10 @@ function supportOpenIsCurrent(generation: number): boolean {
   return generation === supportLifecycleGeneration && initialized && activeIdentity !== null;
 }
 
+function findPostHogButton(selector: string): HTMLButtonElement | null {
+  return document.getElementById(POSTHOG_WIDGET_ID)?.querySelector<HTMLButtonElement>(selector) ?? null;
+}
+
 async function waitForSupportReady(generation: number): Promise<boolean> {
   const deadline = Date.now() + SUPPORT_OPEN_TIMEOUT_MS;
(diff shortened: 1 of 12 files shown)


HamedMP:
@greptile review

PR #1500, opened 2026-08-31
fix(chat): remove legacy shell popover

HamedMP:
## Summary

- remove the legacy shell `ChatPopover` that auto-opened on the Chat app's busy transition
- remove its duplicate desktop/mobile dock triggers, drag-position persistence, animations, Aoede dismissal coupling, and obsolete tests
- retain the canonical Chat app across Web Desktop and Web Canvas
- restore and retain PostHog Conversations customer support on Web and Electron exactly as it exists on `main`
- add a regression guard that prevents the duplicate shell popover from returning

## Tests

- focused Vitest suite covering legacy-overlay removal, canonical Chat launch, PostHog Conversations, shell replay privacy, and Desktop release configuration — 63 passed
- `./node_modules/.bin/tsc --noEmit -p shell/tsconfig.json` — passed
- `./node_modules/.bin/tsc --noEmit -p desktop/tsconfig.json` — passed
- `bun run check:patterns` — 0 violations
- `bun run typecheck` was attempted, but the linked worktree's pnpm build stage stalled; the directly affected shell and Desktop TypeScript projects pass
- the repository-wide Vitest suite was previously attempted in this linked worktree and encountered unrelated environment failures in CLI subprocess, native memory/IPC, host-script, and home-mirror/runtime suites; all affected focused suites pass

## Invariants

- Source of truth: the built-in `__chat__` app is the only Matrix conversation surface; PostHog Conversations remains the separate customer-support surface.
- Lock/transaction scope: none; this is client UI composition only.
- Acceptable orphan states: the old `matrix:chat-popover-offset` localStorage value may remain inert on existing clients and is never read again.
- Auth source of truth: unchanged.
- Deferred scope: none.


HamedMP:
@greptileai Please re-review the current head (2e65be1cc). The prior review covered the superseded PostHog-removal diff; the current PR restores PostHog Conversations and only removes the legacy Matrix ChatPopover.

PR #1538, opened 2026-09-04
feat(cli): add remote computer MCP

HamedMP:
## Summary

- add a stdio Matrix MCP server to the published CLI with 15 tools for computer discovery, captured command execution, persistent zellij sessions and tabs, bounded file transfer, and read-only chat inspection
- return stable Zellij tab IDs atomically and add backward-compatible stable-ID selection while preserving position-based shell clients
- resolve tab working directories within Matrix home, preserve actionable authentication errors, and exit cleanly after MCP stdio disconnects
- ship the MCP alongside the existing Matrix OS plugin and update its skills to prefer remote MCP tools with CLI fallback
- add the complete Spec Kit plan and companion public documentation

Companion docs: https://github.com/FinnaAI/matrix-os-site/pull/73

## Invariants

- source of truth: the existing platform computer inventory and per-computer gateway remain authoritative; the MCP process does not replicate remote state
- auth source of truth: the selected Matrix CLI profile supplies the owner principal, and non-active computers receive only a bounded in-memory runtime-selection token
- lock and transaction scope: no database write, lock, or transaction is added; Zellij emits the created tab ID from the same process invocation, avoiding list/create inference or a new lock
- acceptable orphan states: a created zellij session or tab, or an already-started captured command, may survive an MCP client disconnect and remains discoverable through Matrix; no MCP-local durable state is orphaned
- every scoped tool requires an explicit owner-authorized Matrix computer
- credentials stay in the Matrix CLI profile store and are never embedded in plugin configuration or tool output
- authenticated remote requests reject redirects, use explicit timeouts, and bound responses; client-visible failures use a fixed safe error allowlist
- file tools accept only normalized Matrix-home-relative paths and never read or write coding-agent-local paths
- chat tools are GET-only; no new persistence layer or database migration is introduced
- the one new terminal subroute uses the existing computer-scoped gateway auth boundary, bounded-body middleware, and validated session/tab IDs
- MCP cannot disable host-provided local tools; strict remote-only operation remains a coding-agent host policy
- the final diff is 2,991 additions across 40 files, within the repository's single-PR review band

## Validation

- 33 focused MCP unit and integration tests pass, including a real subprocess handshake and clean shutdown
- the real terminal route-to-adapter test passes for relative/default cwd and rejects missing, traversal, and out-of-home symlink paths
- the broader terminal/plugin suite passes 56/57 tests; its existing shell-wrapper test hits its 1-second timeout after producing the expected stdout, including in isolation, unrelated to tab creation
- both plugin contract tests pass
- sync-client and gateway TypeScript builds pass
- sync-client pre-publish check and plugin validator pass
- all underlying repository typecheck workspace compilers/builds pass when run directly
- changed-file pattern scan reports zero violations
- companion docs suite passes 50 tests

The repository-wide test command was exercised but cannot complete cleanly in this checkout Node 22 shell: Matrix OS requires Node 24, while the installed QMD better-sqlite3 binary targets Node 24 ABI 137. All feature-focused tests pass. The one sync-client credential-discovery failure is environment-sensitive and reproduces on main because this machine has a Claude Code Keychain credential.

## Deferred scope

- no generic operating-system process registry; use captured `run_command` or persistent zellij controls
- no large-file streaming beyond the documented 256 KiB text and 1 MiB binary caps
- no chat mutation
- persistent terminal output is observed through Matrix terminal surfaces; captured output uses `run_command`

### DIFF
--- a/.specify/feature.json
+++ b/.specify/feature.json
@@ -1,3 +1,3 @@
 {
-  "feature_directory": "specs/118-fix-terminal-clipboard"
+  "feature_directory": "specs/120-remote-computer-mcp"
 }
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -477,6 +477,8 @@ Read these on demand, not every session:
 - N/A; selection, context-menu snapshots, and clipboard operation state are transient and must not be persisted (118-fix-terminal-clipboard)
 - Node.js 24+, TypeScript 5.5+ strict, ES modules + Claude Agent SDK V1 `query()`/`resume`, Hono, Zod 4 (`zod/v4`), Kysely, PostgreSQL, Next.js 16, React 19, Cloudflare AI Gateway, OpenRouter OAuth/API (118-ai-gateway-provider-auth)
 - owner configuration/secret files for provider credentials; owner Postgres only if provider account orchestration needs durable multi-instance state; platform PostgreSQL/Kysely for later entitlements, reservations, and content-free usage records (118-ai-gateway-provider-auth)
+- TypeScript 5.9 strict, ES modules; Node.js 24 production target (published CLI remains compatible with Node.js 20+) + `@modelcontextprotocol/sdk` 1.29, Zod 4, citty, native Fetch/AbortSignal, existing Matrix CLI shell/file/profile clients (120-remote-computer-mcp)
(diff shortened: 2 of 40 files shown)


HamedMP:
@greptile review

HamedMP:
@greptile review

HamedMP:
@greptile review

PR #1540, opened 2026-09-05
fix(installer): install terminal runtime user units

zaheerm:
## Summary
- install the bundled `matrix-zellij@.service` and `matrix-terminal.slice` files into `/etc/systemd/user` during standalone self-host installation
- enable lingering, start the `matrix` user manager, and reload its unit cache before the gateway starts
- fail the installation if either bundled unit is absent or a symlink

## Verification
- `bash -n scripts/install-server.sh`
- `pnpm exec vitest run tests/scripts/self-host-installer.test.ts tests/deploy/customer-vps/user-systemd-terminal-runtime.test.ts`

## Invariants
- **Source of truth:** The signed Matrix host bundle supplies the two user units from `/opt/matrix/user-systemd`; the installer copies them as root-owned regular files into the gateway's required `/etc/systemd/user` location.
- **Lock/transaction scope:** No shared data or database mutation. Unit installation and systemd manager setup run synchronously before the installer starts `matrix-gateway.service`.
- **Acceptable orphan states:** If installation stops after copying the files but before service startup, the root-owned unit files and enabled linger state are safe, idempotent prerequisites for a later installer retry.
- **Auth source of truth:** Unchanged. This operates only as root within the existing server installer and does not add an endpoint, credential, or authentication path.
- **Deferred scope:** Existing hosts that already ran a defective installer require the documented one-time repair or a future update/reconciliation path; this PR fixes clean and repeat standalone installs.

### DIFF
--- a/scripts/install-server.sh
+++ b/scripts/install-server.sh
@@ -536,6 +536,10 @@ install_terminal_runtime_user_units() {
       || fail "bundled terminal runtime unit is missing or unsafe: $unit"
     install -o root -g root -m 0644 "/opt/matrix/user-systemd/$unit" "/etc/systemd/user/$unit"
   done
+  [ -x /opt/matrix/terminal-runtime/current/matrix-terminal-attach.mjs ] \
+    || fail "bundled terminal runtime attach helper is missing or not executable"
+  install -d -o root -g root -m 0755 /usr/local/bin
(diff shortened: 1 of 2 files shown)


zaheerm:
Addressed both Greptile findings in `470108505`:

- The standalone installer now installs executable `/usr/local/bin/matrix-terminal-attach` from the active immutable terminal-runtime generation before gateway startup.
- The regression test asserts that helper contract and ordering.

Verified with `bash -n scripts/install-server.sh` plus the 20 targeted Vitest tests.

PR #1554, opened 2026-09-06
fix(desktop): use the keychain password during macOS signing

HamedMP:
## Summary

The Electron dev build on main failed twice on macOS ARM with `SecKeychainUnlock` during keychain ACL setup ([failed run](https://github.com/HamedMP/matrix-os/actions/runs/34060861036)). The pinned `app-builder-lib` creates its temporary keychain with a random password but passes the certificate password to `security set-key-partition-list`.

Patch the pinned dependency to carry the generated keychain password into ACL setup. Certificate import continues to use each certificate's own password, and signing and notarization remain enabled. Copy the patch into Docker dependency stages and host bundles, and mount it alongside package metadata in local/dev Compose services so their package metadata remains self-contained. No dependency versions change.

## Tests

- Red/green regression against the installed dependency, intercepting OS calls: application-only and application-plus-installer certificates.
- Red/green packaging checks for both Docker dependency stages and host bundle patch inclusion.
- Frozen-lockfile install passed.
- 102 initial focused signing, release workflow, and host-bundle tests passed; 21 focused Compose, dev-entrypoint, and signing tests passed after adding the missing dev-container mount.
- Signed desktop builds passed on Linux x64, macOS arm64, and macOS x64, including macOS notarization and DMG launch smoke tests ([artifact run](https://github.com/HamedMP/matrix-os/actions/runs/34064680433)).
- All four Linux unit shards and E2E passed on the first revision. Docker smoke exposed missing patch mounts in Compose; these are now covered by regression tests.
- Repository typecheck passed; pattern checks reported zero violations.
- Full local macOS suite: 13,581 passed, 11 failed, 2 skipped. Failures are in unchanged Linux host-script tests (GNU stat / add-apt-repository), committed DMG artwork byte comparison, and interactive Bash timeout tests. Linux CI and signed macOS packaging remain the release gates.

## Invariants

- Source of truth: the random password used to create the temporary keychain also unlocks it and configures its ACL. Certificate passwords only unlock their corresponding certificate imports.
- Lock/transaction scope: no database or concurrency behavior changes; existing temporary-keychain lifecycle is preserved.
- Acceptable orphan states: none added; existing builder cleanup remains responsible for temporary keychains.
- Auth source of truth: existing CI signing certificates and credentials; no signing, notarization, or credential checks are bypassed.
- Deferred scope: remove the patch when an upstream upgrade passes the regression tests and signed macOS builds on both architectures, as documented in `patches/README.md`.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -16,6 +16,7 @@ WORKDIR /app
 
 # Copy only dependency manifests -- changes here bust the install cache
 COPY package.json pnpm-lock.yaml pnpm-workspace.yaml tsconfig.json ./
+COPY patches/ patches/
 COPY packages/kernel/package.json packages/kernel/
 COPY packages/gateway/package.json packages/gateway/
 COPY packages/observability/package.json packages/observability/
--- a/Dockerfile.platform
+++ b/Dockerfile.platform
@@ -6,6 +6,7 @@ RUN corepack enable && corepack prepare pnpm@10.33.4 --activate
 WORKDIR /app
 
 COPY package.json pnpm-lock.yaml pnpm-workspace.yaml tsconfig.json ./
+COPY patches/ patches/
 COPY packages/clerk-sync/package.json packages/clerk-sync/package.json
 COPY packages/contracts/package.json packages/contracts/package.json
 COPY packages/gateway/package.json packages/gateway/package.json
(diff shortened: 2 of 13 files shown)


HamedMP:
[codex] @greptile review. Commit bb359ef31 mounts the dependency patches alongside package metadata in dev/staging Compose services after Docker smoke exposed the missing mount. Regression tests cover every Compose configuration; 21 focused tests pass. The signing patch passed signed and notarized builds plus DMG launch checks on both Mac architectures.

PR #1555, opened 2026-09-07
fix(billing): preserve portal access across runtime plans and clients

HamedMP:
## Summary

Runtime entitlement overrides incorrectly disabled the billing portal for accounts that already had a Stripe customer. Portal eligibility now comes from the signed-in account's persisted billing customer linkage, matching the existing authenticated portal endpoint. The existing response shape is preserved so deployed clients can consume the fix.

Web Desktop and Web Canvas retain invoice/payment access for overdue, canceled, and unpaid subscriptions instead of showing checkout alone. Missing billing linkage no longer labels paying accounts internally managed. Electron Desktop always shows its billing management entry point, explains unavailable states, and opens eligible portal sessions in the browser.

Native Mobile now consumes the canonical public billing-status contract and follows `portalAvailable` rather than inferring portal access from the effective entitlement source or a private Stripe subscription identifier. This preserves billing management for override-backed accounts that remain linked to a billing customer.

The billing status handler and web billing management components were extracted before adding behavior to their large composition files. The PR-wide trailing-whitespace issue reported by `git diff --check` is also fixed.

## Tests

- 9 focused Native Mobile settings/request tests pass, including override-backed portal access and the no-portal pricing fallback.
- Root `bun run typecheck` passes.
- Pattern scan passes with 0 violations and 5 existing warnings.
- `git diff --check` passes across the PR diff.
- The root `bun run test` run was attempted but stopped after unrelated local-environment failures in coding-agent review snapshots, default apps, host scripts, and app-server runtime tests. The changed billing tests remained green.
- 160 previously validated relevant tests cover platform routes/entitlements, public contracts, customer/override and owner-isolation regressions, web billing flows, and Electron billing management.
- 3 public billing documentation checks pass.
- Local interactive Electron component preview was verified with sample data. Live customer records and production portal sessions have not been verified.

## Invariants

- Source of truth: owner-bound `billing_customers` lookup for portal capability; effective entitlement continues to govern runtime access independently.
- Lock/transaction scope: one additional owner-scoped read; no database writes or schema changes introduced.
- Acceptable orphan states: a portal session can exist after browser-launch failure; retries remain available.
- Auth source of truth: existing resolved Clerk/native Matrix session. No customer IDs are accepted from clients or added to public responses.
- Deferred scope: no Stripe subscription changes, customer data repair, or deployment. Companion public documentation PR: https://github.com/FinnaAI/matrix-os-site/pull/85. Accounts with no effective entitlement still return `entitlement: null` in the existing status contract; the fix targets existing runtime entitlements and does not change that contract.

### DIFF
--- a/apps/mobile/__tests__/billing-settings-screen.test.tsx
+++ b/apps/mobile/__tests__/billing-settings-screen.test.tsx
@@ -1,0 +1,66 @@
+const mockUseSettingsBilling = jest.fn();
+const mockOpenPortal = jest.fn();
+
+jest.mock("@/lib/queries/use-settings-billing", () => ({
+  useSettingsBilling: () => mockUseSettingsBilling(),
+}));
+
+import React from "react";
+import { fireEvent, render, screen, waitFor } from "@testing-library/react-native";
+import { Linking } from "react-native";
+
+import BillingSettingsScreen from "../app/settings-detail/billing";
+
+const overrideEntitlement = {
+  source: "override",
+  planSlug: "matrix_builder",
+  status: "active",
+  portalAvailable: true,
+  billingInterval: null,
+};
+
+function settingsBillingState(portalAvailable: boolean) {
+  return {
+    billing: {
+      entitlement: { ...overrideEntitlement, portalAvailable },
+    },
+    isPending: false,
+    isError: false,
+    openPortal: mockOpenPortal,
+    isOpeningPortal: false,
+  };
+}
+
+describe("native mobile billing settings", () => {
+  beforeEach(() => {
+    jest.clearAllMocks();
+    mockOpenPortal.mockResolvedValue("https://billing.stripe.test/session");
+    mockUseSettingsBilling.mockReturnValue(settingsBillingState(true));
+  });
+
+  it("opens billing management for an override-backed account with a linked customer", async () => {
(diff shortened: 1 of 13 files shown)


Nima-Naderi (changes requested):
Needs changes:
[P1] Native Mobile still uses the obsolete portal rule.
The mobile screen checks source === "stripe" and stripeSubscriptionId, so an override-backed account with a linked billing customer is sent to pricing instead of the portal. Worse, its local status schema requires stripeSubscriptionId, which the public contract intentionally omits.
Evidence: [mobile screen](https://github.com/HamedMP/matrix-os/blob/db4ab9921709e6f46f79ab3091856d3e9d5db162/apps/mobile/app/settings-detail/billing.tsx#L46-L53), [mobile schema](https://github.com/HamedMP/matrix-os/blob/db4ab9921709e6f46f79ab3091856d3e9d5db162/apps/mobile/lib/requests/settings.ts#L14-L25), [public contract](https://github.com/HamedMP/matrix-os/blob/db4ab9921709e6f46f79ab3091856d3e9d5db162/packages/contracts/src/billing-public.ts#L8-L36).
Fix: consume MatrixBillingStatusSchema, gate on entitlement.portalAvailable, and add Native Mobile tests for override/customer and unavailable states. This is required by the repository’s surface-parity rule.

Nima-Naderi (approved):
All good now 👍

PR #1572, opened 2026-09-07
feat(collaboration): preserve actors across platform transport

Nima-Naderi:
## Summary

- Add the content-free platform collaboration directory, discovery index, rollout policy, and one-use ticket persistence.
- Bind short-lived signed actor proofs to actor, audience, method, exact path/query, body digest, policy revision, and key ID.
- Add exact-route proxy parsing, caller credential stripping, bounded upstream responses, safe failures, and API timeouts.

Stack: 2/8 of PR1/M1, based on #1571 and planning PR #1558. This is a review split required by repository size limits.

## Tests

- Platform collaboration repository, proof, and proxy tests are included in the cumulative focused suite: 148 passed.
- Changed TypeScript packages passed their pnpm-equivalent checks.
- `pnpm run check:patterns` — 0 violations.

## Invariants

- **Source of truth:** The owner runtime remains authoritative for grants and content. Platform Postgres stores routing/discovery metadata only.
- **Lock/transaction scope:** Directory/index/policy/ticket mutations are local database operations. Runtime network forwarding is outside those transactions and never turns directory state into authorization.
- **Acceptable orphan states:** Platform metadata may be stale or temporarily unavailable; the owner runtime must still deny unauthorized actors. Failed proxy calls return generic unavailable responses.
- **Auth source of truth:** The authenticated platform actor is preserved in a narrowly scoped signed proof. Caller-supplied proof and owner credentials are stripped.
- **Deferred scope:** Gateway Chat resource authorization, WebSocket forwarding, route registration, and clients arrive in later stack layers. No collaboration mode is enabled here.

## Review/Monitoring

- Draft; review after #1571.
- Greptile and label-gated CI are pending.

PR #1571, opened 2026-09-07
feat(collaboration): add shared authority

Nima-Naderi:
## Summary

- Define strict Zod 4 contracts for collaboration scopes, roles, invitations, membership, events, lifecycle, policy, and actor proofs.
- Add additive owner-Postgres authority tables and canonical Chat attribution fields.
- Implement transactional direct-scope, membership, idempotency, audit, event, and directory-outbox persistence with the complete owner/editor/viewer authority matrix.

Stack: 1/8 of PR1/M1, based on planning PR #1558. This is a review split required by repository size limits, not an additional product milestone.

## Tests

- Cumulative PR1 suite: `pnpm exec vitest run collaboration chat-sharing` — 148 passed; opt-in real-Postgres tests skip without `MATRIX_TEST_POSTGRES_URL`.
- `pnpm --filter '@matrix-os/gateway' exec tsc --noEmit`
- `pnpm run check:patterns` — 0 violations; inherited warning groups only.

## Invariants

- **Source of truth:** Live grants, roles, operations, audit, and Chat attribution live in the content owner's Postgres database. Snapshot tables and tokens remain separate.
- **Lock/transaction scope:** Scope creation uses `ON CONFLICT`; every membership mutation locks the direct scope first and commits the membership, conditional revision, audit, event, operation, and directory outbox together. No network call occurs in those transactions.
- **Acceptable orphan states:** A committed directory-outbox record may remain undelivered until retry; missing or stale platform metadata never grants access.
- **Auth source of truth:** Current accepted owner-local membership and scope lifecycle authorize operations. There is no participant-to-owner impersonation or configured-owner fallback.
- **Deferred scope:** HTTP/WS wiring, platform ingress, clients, AI execution, terminal sharing, and project sharing are later stack layers/milestones. This layer enables no runtime capability.

## Review/Monitoring

- Draft while the stack is reviewed bottom-up.
- Greptile and label-gated CI are pending; no merge or enablement is requested.

PR #1558, opened 2026-09-07
docs(spec): define and phase collaboration session sharing

Nima-Naderi:
## Summary

Define collaboration for an entire project or one standalone Chat/terminal,
delivered through **six implementation PRs across four usable milestones**.
Combine contracts, backend, UI and tests where reviewable to reduce CI
overhead.

| Planned PR | Scope | Result |
| --- | --- | --- |
| PR1 | Common sharing foundation and complete Chat discussion, reusing #1551 | M1: share history and discuss; AI disabled |
| PR2 | Prove and build isolated execution | Disabled execution foundation |
| PR3 | Shared AI queue, controls and UI | M2: prompt AI together |
| PR4 | Terminal sharing, control and UI | M3: use the same eligible terminal |
| PR5 | Project inventory, resource adapters and migration | Disabled project foundation |
| PR6 | Project sharing UI and full integration | M4: share an entire project |

Each PR includes its own tests. PR2 and PR5 stay disabled until their consumer
milestones complete. Split further only with actual reviewability or
repository-limit evidence. Four documentation updates in separate
`FinnaAI/matrix-os-site` PRs accompany releases without gating implementation
merges or internal enablement. These are planning IDs; no implementation PRs or
deployments are created here.

`tasks.md` now expands the accepted plan into 125 dependency-ordered, test-first
tasks organized by user story and mapped back into exactly these six PRs. It
records independent test criteria, parallel opportunities, the fixed Graphite
stack, milestone validation, screenshots, rollback drills, and four separate
site-documentation updates. Individual tasks remain implementation units, not
new PR boundaries.

## Reuse of Merged Snapshot Sharing

Rebased onto main at `7e333712e8914d173f5c023aae7d751a551de928`
(merged #1551). The existing common Share button/dialog and Web/Electron
adapters become PR1's entrypoint, with **Share snapshot** and **Invite
collaborators** as distinct actions. Preserve snapshot preview/renewed consent,
seven-day expiry, owner revoke, safe Markdown and excluded
attachments/tool/hidden context. Reuse canonical attachment/navigation
improvements with scope-authorized destinations.

Snapshot storage, reduced schema and anonymous bearer relay remain separate
from canonical live history, participant identity and current membership
checks. A snapshot token never grants live HTTP/WS access or execution;
collaborator status never grants snapshot-management rights. Link revocation
and membership revocation affect their respective targets. PR1 includes
existing snapshot regressions and tests of the two-action choice, credential
separation and independent revocation. This reduces PR1 UI work without
[...]

PR #1585, opened 2026-09-08
revert(collaboration): undo combined stack squash

Nima-Naderi:
## Summary

- Revert the accidental squash merge from PR #1572 that combined the planning layer and the first two implementation layers into one `main` commit.
- Restore `main` exactly to the tree before commit `45c0708572e7c074f2a60369ca9774fa24cb5afd` so the stack can be re-landed one PR at a time.
- No collaboration behavior is intended to remain on `main` after this recovery PR; each reviewed layer will be restored through its own replacement PR.

## Tests

- `git diff --quiet 45c0708572e7c074f2a60369ca9774fa24cb5afd^ -- .` (pass; resulting tree exactly matches the pre-squash parent)
- `pnpm run check:patterns` (pass; 0 violations, 5 existing warnings)
- `pnpm run typecheck` / `pnpm run test` could not start locally because the fresh manual worktree has no package-local `node_modules`; full CI is required before merge.

## Invariants

- **Source of truth:** `main` before the accidental squash is the recovery target; this revert produces that exact tree.
- **Lock/transaction scope:** no runtime database operation or transaction behavior is introduced; this PR only reverses an accidental Git merge.
- **Acceptable orphan states:** none. The replacement stack will not be merged until this revert is merged and `main` is clean.
- **Auth source of truth:** unchanged from the pre-squash `main` tree.
- **Deferred scope:** re-landing planning, shared authority, actor transport, and later Chat collaboration layers is intentionally handled by subsequent PRs.


Nima-Naderi:
[codex] @greptileai please review the exact recovery head bd3b70ce8. This PR restores main byte-for-byte to the parent of the accidental combined squash.

Nima-Naderi:
Closing this recovery PR because we have decided to retain PR #1572's combined squash as the current main baseline and cleanly restack only the remaining collaboration layers on top of it.\n\nAlthough #1572 unintentionally combined the planning changes from #1558 and the shared-authority changes from #1571 with its own actor-transport layer, reverting it now would introduce unnecessary revert/reapply churn. More importantly, Greptile identified a real rollback hazard: a database that booted the combined squash may retain the new chat_messages.purpose NOT NULL constraint while reverted writers omit purpose, which could break message writes.\n\nTherefore we will not merge this revert. The contents of #1558, #1571, and #1572 will remain on main through #1572's squash commit, and the unique deltas from the later PRs will be rebuilt/restacked from current main. This preserves the working schema transition and avoids risking deployed databases solely to correct PR history.

PR #1597, opened 2026-09-09
fix(vps): complete restore downloads before starting runtime

HamedMP:
## Summary

- Fix restore downloads that exit successfully without publishing the destination: an `autoClose: false` stream retains the file handle, leaving `close()` pending before rename.
- Stream bounded response chunks directly into the owned handle, preserving backpressure, private temporary files, fsync, and atomic rename.
- Reconcile Docker group membership for an existing snapshot user before restore starts.
- Add public-safe operator documentation and subprocess regression tests.

## Tests

- TDD: reproduced missing destination on the original CLI, then green after the fix.
- 76 tests pass across `r2-broker-download`, `snapshot-restore-groups`, and `customer-vps-host-bundle`.
- `node --check` and `git diff --check` pass.
- `bun run check:patterns`: zero violations, five existing warnings.
- Full typecheck/test commands attempted but package-manager bootstrap stalled; stopped those commands. Full-suite pass is not claimed.
- Disposable snapshot-based VPS: restore completed, exact requested bundle installed, gateway/shell/sync active and Web Desktop loaded. No production fleet rollout.

### Source of truth
- Durable owner backups remain in the existing object store. Restore completion and actual runtime health gate readiness; power-on alone does not.

### Lock/transaction scope
- Exclusive temporary-file creation; one file-handle owner; awaited writes, fsync and close before atomic rename. No database changes.

### Acceptable orphan states
- Caught failures remove the temporary download and preserve an existing destination. Abrupt host/process termination may leave a private temporary file, as before.

### Auth source of truth
- Existing broker authorization and scoped presigned downloads are unchanged. Append only the Docker group already declared for the runtime account; do not relax socket permissions.

### Deferred scope
- Loading/error UI, TTL preview retention, and Cloudflare token-count authentication are separate concerns. Nothing in this PR claims funded inference works or requests a merge/deployment.

### DIFF
--- a/distro/customer-vps/cloud-init.yaml
+++ b/distro/customer-vps/cloud-init.yaml
@@ -584,6 +584,9 @@ runcmd:
       exit 1
     fi
     usermod -d /home/matrix/home matrix
+    # Cloud-init does not reapply supplementary groups to an existing snapshot
+    # user. The restore gate must access the local Postgres container as matrix.
+    usermod -aG docker matrix
     install -d -o matrix -g matrix -m 0750 /home/matrix/home /home/matrix/home/projects /home/matrix/projects /var/lib/matrix/db /var/lib/matrix/db/snapshots /var/lib/matrix/logs
     ensure_owner_link() {
       local name="$1" mode="$2" legacy="/home/matrix/$1" target="/home/matrix/home/$1"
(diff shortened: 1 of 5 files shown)


HamedMP:
@greptile review

PR #1623, opened 2026-09-11
fix(billing): validate portal redirects consistently across clients

### LINKED ISSUE #1557: fix(billing): validate portal redirects consistently across clients
## Problem

Billing portal redirect responses are validated inconsistently across clients:

- Electron Desktop requires a valid URL, caps it at 2,048 characters, and requires HTTPS.
- Web checks only that the response contains a truthy URL before passing it to window.location.assign.
- Native Mobile validates general URL syntax but does not require HTTPS.

The platform normally returns a trusted Stripe-hosted URL, so this is defense-in-depth rather than evidence of an active exploit. A malformed or regressed upstream response should nevertheless be unable to navigate a client to a relative, HTTP, or executable-scheme URL.

Context: identified while reviewing #1555.

## Expected behavior

All billing clients should accept only bounded HTTPS portal redirects and show a safe, retryable error for rejected responses without opening a browser or navigating the current page.

## Acceptance criteria

- Define or reuse one shared billing redirect response schema.
- Require a syntactically valid HTTPS URL with a maximum length of 2,048 characters.
- Apply the schema consistently in Web, Electron Desktop, and Native Mobile.
- Reject relative URLs, HTTP URLs, executable schemes such as javascript, missing URLs, and overlong URLs.
- Do not expose provider responses or raw errors in user-facing messages.
- Add regression tests proving every applicable client rejects unsafe redirects and accepts a normal Stripe HTTPS redirect.

## Scope notes

- Do not add a Stripe-host allowlist unless supported Stripe portal/custom-domain behavior has first been verified.
- This issue does not change portal eligibility or billing authorization; those remain server-side and owner-scoped.

shashank-100:
## Summary

Fixes #1557.

Billing portal redirect responses were validated inconsistently across clients:

- **Electron Desktop** required a valid, bounded (2,048 char) HTTPS URL.
- **Web** checked only that the response contained a truthy `url` before passing it to `window.location.assign` — on both the portal and the checkout path.
- **Native Mobile** validated general URL syntax via `z.url()`, which accepts `http://` and `javascript:` and has no length bound.

The platform normally returns a trusted Stripe-hosted URL, so this is defense in depth rather than evidence of an active exploit. A malformed or regressed upstream response should nevertheless be unable to navigate a client to a relative, plaintext, or executable-scheme URL.

## Change

Promote the existing Electron schema into `@matrix-os/contracts` as the single billing redirect contract and apply it in all three clients:

- `MatrixBillingRedirectSchema` — strict object; `url` must be a syntactically valid absolute HTTPS URL of at most `MATRIX_BILLING_REDIRECT_MAX_LENGTH` (2,048) characters.
- `parseBillingRedirectUrl(body)` — parses an untrusted response body and returns the safe URL or `null`.

All four navigating call sites now validate through the shared contract: Web portal (`BillingManagementPanel.tsx`), Web checkout (`BillingCheckoutPanel.tsx`), and both Electron checkout and portal (`BillingSection.tsx`). Rejected responses surface the existing generic, retryable error. No provider response or raw error is exposed to the user. Portal eligibility and billing authorization are unchanged and remain server-side and owner-scoped.

Per the issue's scope note, no Stripe-host allowlist was added, since supported Stripe portal/custom-domain behavior has not been verified here.

## Acceptance criteria

- [x] Define or reuse one shared billing redirect response schema
- [x] Require a syntactically valid HTTPS URL with a maximum length of 2,048 characters
- [x] Apply the schema consistently in Web, Electron Desktop, and Native Mobile
- [x] Reject relative URLs, HTTP URLs, executable schemes such as `javascript:`, missing URLs, and overlong URLs
- [x] Do not expose provider responses or raw errors in user-facing messages
- [x] Add regression tests proving every applicable client rejects unsafe redirects and accepts a normal Stripe HTTPS redirect

## Validation

- `pnpm exec vitest run tests/contracts/billing-public.test.ts tests/shell/billing-portal-access.test.tsx` — 20/20 pass
- `pnpm --dir apps/mobile exec jest __tests__/billing-portal-redirect.test.ts` — 9/9 pass (mobile suite: passing suites 44 -> 45, passing tests 282 -> 291; failures unchanged)
- `pnpm exec vitest run tests/shell/billing-section.test.tsx` — 5 new Web checkout rejection cases pass (33 -> 38 passing; the 3 pre-existing failures in this file are unchanged). Reverting only the checkout fix raises failures from 3 to 8, confirming the new cases are non-vacuous.
- `bun run typecheck` — passes for all packages, including `desktop`
- `bun run check:patterns` — 0 violations (5 pre-existing warnings in untouched files)
- `npx react-doctor@latest shell` and `... desktop` — finding counts byte-identical to baseline in every category

**The new tests were confirmed to actually catch the defect.** Reverting only the Web fix to the previous truthy-`url` check makes 5 of the new cases fail, demonstrating the shell really did navigate to `javascript:`, `data:`, relative, plaintext, and overlong URLs before this change.

### Pre-existing failures (verified against clean `main`, not introduced here)

- 4 failing tests in `tests/shell/billing-gate.test.tsx` / `billing-section.test.tsx`
- 11 `tsc` errors and 14 failing Jest suites under `apps/mobile/` (the documented `pnpm.packageExtensions` workspace-resolution gotcha)

### Native Mobile client-level coverage

`apps/mobile/__tests__/billing-portal-redirect.test.ts` — 9 tests covering the changed Native Mobile request wiring directly:

- a normal Stripe HTTPS redirect is returned
- relative, plaintext, executable-scheme, `data:`, overlong, and missing redirects are all rejected
- **a rejected redirect never reaches `Linking.openURL`** (the billing screen calls `await Linking.openURL(await openPortal())`)
- the surfaced error does not echo the upstream response

The `@matrix-os/contracts` barrel reaches `chat-sharing.ts` -> ESM-only `micromark`, which Jest cannot transform in `apps/mobile`. That is unrelated to billing, so the test re-exports the real `billing-public` schema module directly rather than stubbing the validation under test — the assertions exercise the actual schema.

Reverting only the mobile schema to the previous `z.url()` makes 6 of these 9 fail, including the `Linking.openURL` guard, confirming `javascript:alert(1)` would previously have reached `openURL` on Native Mobile.

## Surface matrix

| Surface | UI | Behavior | State/recovery | Automated tests | Real evidence |
| --- | --- | --- | --- | --- | --- |
| Web Canvas | N/A | pass | pass | pass | N/A |
| Web Desktop | N/A | pass | pass | pass | N/A |
| Electron Desktop | N/A | pass | pass | pass | N/A |
| Web Mobile | N/A | pass | pass | pass | N/A |
| Native Mobile | N/A | pass | pass | pass | N/A |

**UI `N/A` rationale**: no user-visible UI, copy, layout, or styling changes. The only behavioral change on the rejection path reuses the existing generic error string already rendered by each surface, so there is no new visual state to screenshot.

**Real evidence `N/A`**: reproducing this path against a live surface requires the platform to return a malformed redirect, which it does not do in normal operation. The behavior is instead proven executably by the revert check described under Validation.

## Invariants

- **Source of truth**: the platform billing service remains authoritative for portal eligibility and redirect targets. `MatrixBillingRedirectSchema` in `@matrix-os/contracts` is the single source of truth for what shape of redirect a client may act on; no client re-derives that rule.
- **Lock/transaction scope**: none. This is client-side response validation with no database writes and no new network calls.
- **Acceptable orphan states**: none introduced. Validation happens after the request completes and before navigation, so a rejected redirect leaves the client exactly as it was before the action, with the action re-runnable.
- **Auth source of truth**: unchanged. The portal endpoint stays owner-scoped and server-authorized; this change never widens who may open a portal, only which redirect targets are honored.
- **Deferred scope**: no Stripe-host allowlist (see the issue's scope note); no change to portal eligibility, billing authorization, or checkout flow semantics.

## Deployment note

Per `CLAUDE.md`, the billing-locked Settings UI is served by the platform-owned `app.matrix-os.com` shell surface before a user has an active VPS. Reaching users on that pre-VPS path requires a platform/app-shell deployment; a host-bundle publish alone will not update it.

### DIFF
--- a/apps/mobile/__tests__/billing-portal-redirect.test.ts
+++ b/apps/mobile/__tests__/billing-portal-redirect.test.ts
@@ -1,0 +1,78 @@
+jest.mock("@/lib/storage", () => ({
+  HOSTED_GATEWAY_URL: "https://app.matrix-os.com",
+}));
+
+// The @matrix-os/contracts barrel reaches chat-sharing.ts -> ESM-only micromark,
+// which Jest cannot transform here. That is unrelated to billing, so re-export
+// the real billing schema module directly rather than stubbing the validation
+// under test.
+jest.mock("@matrix-os/contracts", () => require("../../../packages/contracts/src/billing-public"));
+
+import { Linking } from "react-native";
+
+import { createMobileBillingPortal } from "@/lib/requests/settings";
+
+const portalResponse = (body: unknown) => ({
+  ok: true,
+  json: jest.fn().mockResolvedValue(body),
+}) as unknown as Response;
+
+describe("Native Mobile billing portal redirect", () => {
+  beforeEach(() => {
+    jest.restoreAllMocks();
+  });
+
+  it("returns a normal Stripe HTTPS redirect", async () => {
+    const url = "https://billing.stripe.com/p/session/live_abc123";
+    jest.spyOn(global, "fetch").mockResolvedValue(portalResponse({ url }));
+
+    await expect(createMobileBillingPortal("clerk-token")).resolves.toBe(url);
+  });
+
+  it.each([
+    ["a relative path", "/billing/portal"],
+    ["a plaintext URL", "http://billing.stripe.com/p/session"],
+    // eslint-disable-next-line no-script-url
+    ["an executable scheme", "javascript:alert(1)"],
+    ["a data URL", "data:text/html,<script>alert(1)</script>"],
+    ["an overlong URL", `https://billing.stripe.com/p/${"a".repeat(2048)}`],
+  ])("rejects %s instead of returning it", async (_label, url) => {
+    jest.spyOn(global, "fetch").mockResolvedValue(portalResponse({ url }));
+
+    await expect(createMobileBillingPortal("clerk-token")).rejects.toThrow();
+  });
+
+  it("rejects a missing redirect URL", async () => {
+    jest.spyOn(global, "fetch").mockResolvedValue(portalResponse({}));
+
+    await expect(createMobileBillingPortal("clerk-token")).rejects.toThrow();
+  });
+
+  // The billing screen calls `await Linking.openURL(await openPortal())`, so a
+  // rejected redirect must reject before openURL is ever reached.
+  it("never passes an unsafe redirect to Linking.openURL", async () => {
+    const openURL = jest.spyOn(Linking, "openURL").mockResolvedValue(undefined as never);
+    jest.spyOn(global, "fetch").mockResolvedValue(
+      // eslint-disable-next-line no-script-url
+      portalResponse({ url: "javascript:alert(1)" }),
+    );
+
+    await expect(
+      (async () => {
+        await Linking.openURL(await createMobileBillingPortal("clerk-token"));
+      })(),
+    ).rejects.toThrow();
+    expect(openURL).not.toHaveBeenCalled();
+  });
+
+  it("does not expose the upstream response in the surfaced error", async () => {
+    jest.spyOn(global, "fetch").mockResolvedValue(
+      // eslint-disable-next-line no-script-url
+      portalResponse({ url: "javascript:alert(1)" }),
+    );
+
+    await expect(createMobileBillingPortal("clerk-token")).rejects.not.toThrow(
+      /javascript:|stripe|billing\.stripe/i,
+    );
+  });
+});
--- a/apps/mobile/__tests__/billing-portal-redirect.test.ts
+++ b/apps/mobile/__tests__/billing-portal-redirect.test.ts
@@ -71,8 +71,10 @@ describe("Native Mobile billing portal redirect", () => {
       portalResponse({ url: "javascript:alert(1)" }),
     );
 
-    await expect(createMobileBillingPortal("clerk-token")).rejects.not.toThrow(
-      /javascript:|stripe|billing\.stripe/i,
+    // Assert the exact generic message rather than the absence of a pattern:
+    // a "does not match" assertion also passes for an unrelated rejection.
+    await expect(createMobileBillingPortal("clerk-token")).rejects.toThrow(
+      "Billing portal unavailable. Try again.",
     );
   });
 });
--- a/apps/mobile/lib/requests/settings.ts
+++ b/apps/mobile/lib/requests/settings.ts
@@ -17,7 +17,6 @@ const SystemInfoSchema = z.object({
 }).passthrough();
 
 const OkResponseSchema = z.object({ ok: z.literal(true) }).passthrough();
-const BillingPortalSchema = MatrixBillingRedirectSchema;
 
 export type MobileSystemInfo = z.infer<typeof SystemInfoSchema>;
 export type MobileBillingStatus = MatrixBillingStatus;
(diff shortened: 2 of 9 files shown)


shashank-100:
Valid — fixed in dddc4b240 rather than deferred.

I had assumed the mobile Jest suite simply could not import `@matrix-os/contracts`. Re-checking, the actual failure is narrower: the contracts barrel reaches `chat-sharing.ts` -> ESM-only `micromark`, which Jest cannot transform. That is unrelated to billing, so mocking the barrel with the real `billing-public` module gives genuine client-level coverage without a config change or a deferral.

Added `apps/mobile/__tests__/billing-portal-redirect.test.ts` (9 tests), covering exactly the three gaps you named:

- **changed request wiring**: `createMobileBillingPortal` is exercised directly, asserting a normal Stripe HTTPS redirect is returned and that relative, plaintext, executable-scheme, `data:`, overlong, and missing redirects all reject.
- **rejected values never reach `Linking.openURL`**: the billing screen calls `await Linking.openURL(await openPortal())`, so the test drives that same expression with a spy and asserts `openURL` is never called.
- **generic error mapping**: asserts the surfaced error does not echo the upstream response (no `javascript:`, no provider host).

The mock re-exports the real schema module, so these assertions run against actual validation rather than a stub. Verified non-vacuous: reverting only the mobile schema to the previous `z.url()` makes 6 of the 9 fail, including the `Linking.openURL` guard — confirming `javascript:alert(1)` would previously have reached `openURL` on Native Mobile.

Mobile suite: passing suites 44 -> 45, passing tests 282 -> 291 (+9). The 14 pre-existing suite failures and 11 `tsc` errors are unchanged and fail identically on clean `main`. Surface matrix and PR body updated — Native Mobile automated tests now `pass`.

shashank-100:
Finding 1 — valid and fixed in 12cd12592.

You were right, and it undercut the PR's own acceptance criterion. I fixed Web portal but missed Web checkout in the sibling `BillingCheckoutPanel.tsx`, while Electron got both of its call sites — so "apply the schema consistently in Web" was not actually met, and `{url: "javascript:alert(1)"}` would still have reached `window.location.assign` on the checkout path.

`BillingCheckoutPanel.tsx:178` now routes through `parseBillingRedirectUrl`, so all four navigating call sites share one contract: Web portal, Web checkout, Electron portal, Electron checkout.

Added 5 rejection cases to `tests/shell/billing-section.test.tsx`, reusing the existing `onCheckoutNavigate` harness to assert unsafe targets never reach navigation (relative, plaintext, executable-scheme, `data:`, overlong). Verified non-vacuous: reverting only the fix raises failures in that file from 3 to 8. The 3 remaining failures are pre-existing and fail identically on clean `main`.

Gates re-run: contracts + portal 20/20, mobile 9/9, `bun run typecheck` passes, `check:patterns` 0 violations, and react-doctor counts on `shell` are byte-identical to baseline in every category.

Findings 2 and 3 I have not actioned, and I think they are worth separating:

- **2 (competing validators)** — fair hit on the rationale. `ai-credit-checkout.ts` does pin `hostname === "checkout.stripe.com"`, so the claim that Stripe host behavior is unverified is weaker than I wrote. The issue explicitly scoped the allowlist out, so I would rather not widen this PR; better as a follow-up that reconciles both validators deliberately, since the portal and checkout hosts differ (`billing.stripe.com` vs `checkout.stripe.com`) and custom domains are in play.
- **3 (return the normalized `href`)** — agreed this is the right place to tighten, and it is a one-line `.transform()`. Holding it only because it changes the value handed to `location.assign`/`openURL` and deserves its own test rather than riding along here.

Happy to fold either into this PR instead if you prefer.

PR #1628, opened 2026-09-11
fix(terminal): make migration updates recoverable

Nima-Naderi:
## Summary

- give the system terminal controller, Zellij commands, subscriptions, and PTY attachments one explicit owner runtime environment derived from the runtime UID
- prepare every replacement workspace before stopping legacy sessions, and propagate non-missing Zellij deletion failures
- keep gateway/shell available when the terminal control plane is unavailable while requiring terminal health before a release commits
- make host updates recoverable across the app, host helpers, system/user units, unit state, terminal generation pointer, migration journal, and release metadata
- run candidate migration rollback before removing candidate code and recover interrupted update transactions on restart

## Tests

- `pnpm install --frozen-lockfile`
- package-equivalent full TypeScript check: pass (Bun is unavailable in this environment, so the `pnpm` subcommands behind `bun run typecheck` were run directly)
- `pnpm run check:patterns`: pass, 0 violations
- changed-area terminal/deployment suite: 142 tests pass
- user-systemd/environment suite: 73 tests pass
- `bash -n` for the updater and production acceptance script: pass
- `git diff --check`: pass
- repository-wide `pnpm run test`: run locally; unrelated high-contention suites timed out in the monolithic invocation, while the changed-area suites pass and CI shards remain authoritative

## Review/Monitoring

- branch is frozen at the current commit while CI and Greptile review it
- no production or customer VPS deployment is performed by this PR
- after merge, publish the immutable host bundle, deploy that exact version to one disposable/exact-version canary, then recover the two affected VPSes before broader promotion

## Updater extraction plan

`distro/customer-vps/host-bin/matrix-sync-agent` is 1,745 lines after this urgent recovery change. Per `docs/dev/large-file-refactoring.md`, the follow-up will be a separate mechanical extraction so recovery semantics are not changed at the same time as code is moved:

1. Add a domain-named `matrix-update-transaction-lib` under `distro/customer-vps/host-bin/`, installed and backed up by the existing host-bin transaction path.
2. Move only transaction snapshot, validation, restore, enablement/activity reconciliation, and cleanup functions into that sourced helper. Keep polling, download, update orchestration, and service health decisions in `matrix-sync-agent`.
3. Resolve the helper from the fixed `/opt/matrix/bin` install boundary, validate that it is a root-owned regular file before sourcing it, and include the helper in interrupted-update rollback so code and transaction semantics cannot diverge.
4. Mechanically split the transaction assertions into `tests/deploy/customer-vps/update-transaction.test.ts`, preserving behavior and assertion intent, then rerun the two-bundle production acceptance matrix.
5. Ship the extraction as its own lower-risk PR immediately after this incident fix; target `matrix-sync-agent` below 1,500 lines without mixing further updater behavior changes into that PR.

### Source of truth

- owner terminal state and the migration journal under `$MATRIX_HOME/system` remain canonical for workspace identity and migration progress
- `/opt/matrix/release.json` remains the installed release authority; the durable update transaction records only compensation data until health verification commits the candidate
- each user-systemd descriptor continues to pin its immutable terminal generation

### Lock/transaction scope

- the updater records and seals all replaced host artifacts plus systemd enablement/activity before stopping services or mutating the installation
- migration stages and prepares replacement workspaces before crossing the accepted legacy-process interruption boundary
- bundle download and extraction remain outside the mutation transaction; compensation starts only after a validated candidate is local

### Acceptable orphan states

- a preparation failure can leave unused replacement workspace shells, but legacy processes remain alive and journaled rollback removes the replacements
- interruption after mutation leaves a root-owned transaction that the restarted sync agent either finalizes after a proven metadata commit or compensates idempotently
- failed candidate artifacts may remain under bounded staging cleanup after rollback; owner data is never removed

### Auth source of truth

- no request-auth boundary changes; existing gateway principal and VPS upgrade-token contracts remain authoritative
- the local terminal control socket remains owner-only and terminal failures are exposed as generic unavailable responses

### Deferred scope

- already-terminated legacy process memory cannot be reconstructed; this prevents future premature termination and restores service availability, but cannot resurrect a process killed by the prior rollout
- the updater transaction helper extraction is deliberately mechanical follow-up work as specified above; it is separated from this incident fix to avoid combining code movement with rollback semantic changes
- live exact-version canary, affected-VPS recovery, and fleet promotion occur only after merge and immutable bundle publication


Nima-Naderi:
@greptileai please review this terminal migration and updater recovery fix.

Nima-Naderi:
@greptileai please re-review the current head 79aa75283. Both prior findings are addressed: the committed-update transaction now remains until consumed markers are cleared, with a focused regression test, and the PR body now contains the concrete updater extraction plan required by the repository policy.

PR #1646, opened 2026-09-12
test(deploy): smoke stable to current dev upgrade

Nima-Naderi:
Temporary operator-only harness for one released-bundle smoke test. It is intentionally restricted to v2026.09.12-1237 and merge SHA 501faacbecf5f9726c92fa6ef10ad8f89b42a587. Do not merge; close after the disposable VPS test.

## Invariants

- Source of truth: public stable/dev channel manifests plus platform fleet state.
- Lock/transaction scope: one pr-number-scoped preview workflow concurrency group.
- Acceptable orphan states: only the named disposable pr-* VPS; teardown is required after the run.
- Auth source of truth: existing GitHub repository secrets and per-runtime HMAC token.
- Deferred scope: no customer deployment, channel promotion, or production data mutation.

### DIFF
--- a/.github/workflows/preview-vps.yml
+++ b/.github/workflows/preview-vps.yml
@@ -1112,8 +1112,15 @@ jobs:
               "currentStart": current.get("ExecMainStartTimestamp", "unknown")[:64],
           }, sort_keys=True))
           PYTHON
-          body="$(jq -cn --arg script "$lifecycle_script" \
-            '{command:["/usr/bin/sudo","/usr/bin/python3","-c",$script],timeoutMs:30000}')"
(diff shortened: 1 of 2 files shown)


Nima-Naderi:
Test-only draft; do not merge. Main CI and host-bundle release passed for v2026.09.12-1237. Two completed stable-to-dev runs proved terminal migration committed with two seeded legacy sessions converted to two tabs under one active workspace-scoped Zellij service. Both also exposed a transient post-health 502 caused by apt-daily-upgrade re-executing systemd and restarting Matrix services on fresh hosts, not by migration/OOM. A final clean rerun was attempted on two fresh preview handles, but both remained in platform provisioning for 10+ minutes and never attempted the bundle. All disposable VPS records were torn down.

PR #1648, opened 2026-09-12
fix(platform): bootstrap previews from exact bundles

HamedMP:
## Summary
- allow the trusted preview provision route to select an already-published immutable host bundle
- boot fresh preview VPSes from the exact PR bundle instead of the frozen stable bundle
- reject unpublished versions and ambiguous snapshot plus bundle requests before provider mutation

## Tests
- platform TypeScript type-check
- exact published and unpublished preview bundle service tests
- preview provision route auth and validation contract
- preview workflow provision payload contract

## Invariants
- **Source of truth:** platform host bundle release metadata is authoritative; the workflow supplies the immutable version it just published or explicitly validated.
- **Lock/transaction scope:** existing owner-scoped provisioning locks and the machine-claim transaction remain unchanged; release resolution happens before provider creation.
- **Acceptable orphan states:** no new orphan state is introduced; an unknown release fails before Hetzner mutation, while existing cleanup/reconciliation continues to own provider cleanup.
- **Auth source of truth:** only the platform bearer-authenticated preview route may request an exact bundle; snapshot selection remains restricted to the distinct snapshot-operator credential.
- **Deferred scope:** existing previews already stuck on an older bootstrap bundle still require teardown and reprovisioning. Customer provisioning and stable-channel selection are unchanged. No public product documentation changes are needed for this internal preview deployment repair.

### DIFF
--- a/.github/workflows/preview-vps.yml
+++ b/.github/workflows/preview-vps.yml
@@ -332,8 +332,9 @@ jobs:
           provision_body="$(jq -cn \
             --arg owner "$PREVIEW_CLERK_USER_ID" \
             --arg handle "$HANDLE" \
+            --arg bundle_version "$VERSION" \
             --argjson access "$access_user_ids" \
-            '{clerkUserId: $owner, handle: $handle, runtimeSlot: $handle, accessClerkUserIds: $access}')"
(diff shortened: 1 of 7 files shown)


HamedMP:
@greptileai

HamedMP:
@greptileai

PR #1655, opened 2026-09-12
test(desktop): stabilize terminal clipboard E2E for #1650

### LINKED ISSUE #1650: test(desktop): eliminate terminal clipboard E2E timing flake
## Summary

Main CI failed on commit `06bbd789302af4bc62942643453bde1cb5d6abf2` in the required packaged Electron terminal clipboard regression. All unit-test shards, typecheck, pattern scan, React Doctor, shell production build, general E2E tests, and the MAT-335 Desktop regression passed.

Failed run: https://github.com/HamedMP/matrix-os/actions/runs/34689720056

## Failures

Two of seven cases in `tests/e2e/desktop/terminal-clipboard.e2e.test.ts` failed:

1. `selects the complete xterm scrollback rather than only the visible viewport`
   - The one-time right-click did not produce the Terminal actions menu.
   - Playwright timed out for 30 seconds waiting for the exact `Select All` menu item.

2. `extends a drag selection by auto-scrolling beyond both terminal edges`
   - Expected copied text to contain `EDGE-SCROLL-020`.
   - Clipboard still contained `targetword` from the preceding double-click test, showing that the new drag did not establish a selection.

## Assessment

The suite uses fixed 250–300 ms delays after asynchronous terminal output and then performs coordinate-based mouse gestures. Under a slower CI render/layout cycle, the one-time right-click or drag can miss. Tests also share one Electron page, terminal instance, selection state, mouse mode, and clipboard across cases, allowing a failed gesture to expose stale state from a previous test.

The merged Terminal change added binary-input transport and an `onBinary` subscription; it did not change the context-menu or selection implementation. The stub gateway does not advertise the binary-input capability, so this test continues through the legacy input-frame fallback.

This should be treated as an E2E synchronization/state-isolation defect unless a deterministic product regression is demonstrated.

## Release impact

The failure caused the same-SHA host-bundle gate to fail, so build, publish, and deployment were skipped. No bundle from this commit was released.

## Immediate mitigation

Rerun the failed main CI jobs once. If CI passes, separately rerun the failed Host Bundle Release workflow for the same SHA because its completed failed run will not automatically resume.

Do not add a blanket retry or weaken the required assertions.

## Durable fix

- Replace fixed sleeps with readiness checks tied to terminal buffer/render completion and expected marker visibility.
- Verify the intended pointer target is inside the active terminal immediately before each gesture.
- Wait for the context menu to become visible after right-click, with diagnostic evidence if the click misses.
- Reset or assert terminal selection, clipboard contents, viewport position, and mouse-reporting mode between cases.
- Prefer isolated page/session setup for tests whose pointer or terminal modes can leak across cases.
- Assert that a new selection exists before invoking copy so stale clipboard values cannot obscure the actual failure.
- Run the packaged Electron suite repeatedly under Xvfb to verify stability without weakening product behavior checks.

## Acceptance criteria

- The required clipboard suite passes repeatedly under the CI/Xvfb environment.
- Tests remain sensitive to genuine context-menu, scrollback selection, auto-scroll selection, clipboard, and mouse-reporting regressions.
- No fixed-delay-only synchronization for terminal output readiness.
- A failed pointer gesture reports terminal geometry, hit target, viewport/buffer state, selection, and mouse-reporting state.

sundaram2021:
## Summary

Stabilizes the packaged Electron terminal clipboard E2E suite that flaked in main CI: the one-time right-click never produced the Terminal actions menu (`Select All` 30s timeout), and the edge-drag never established a selection (clipboard still held `targetword` from the prior double-click test).

Root cause is E2E synchronization/state-isolation, not a product regression: fixed 250–300ms sleeps after async terminal output, unvalidated coordinate gestures, and shared page/selection/clipboard/mouse-mode state across cases.

## Relevant files

**`tests/e2e/desktop/terminal-clipboard.e2e.test.ts`**
- Replaced all output-readiness `waitForTimeout(250/300)` and the drag `waitForTimeout(1500)` with `waitForTerminalOutput()` (xterm buffer poll) + `waitForRenderFrames()` + `waitForMouseTrackingMode()`.
- New `waitForMarkerAtVisibleRow()`: asserts the marker sits at the expected visible row before computing `point(col,row)`, so surviving scrollback can't mis-aim gestures while the buffer scan still passes.
- New `scrollTerminalToBottom()` after each output burst (including scrollback/edge cases) to pin the viewport before gesturing.
- Every pointer target validated via `expectTerminalPoint()` (hit-test + `isTerminalSurface`), including the previously raw `point(2,0)` resume click.
- Right-click path waits for menu + `Select All` visible, with `readTerminalDiagnostics` dump + screenshot on miss.
- `beforeEach` now: dismisses menu, disables mouse-reporting, clears selection via the new renderer hook, scrolls to bottom, polls selection to `""`, writes a clipboard sentinel.
- Every copy asserts a fresh selection first (`expect.poll(readTerminalSelection)`) so stale clipboard values can't mask failures.
- The one remaining `waitForTimeout(500)` is documented inline as an intentional mouse-stability soak, not readiness sync.

**`desktop/src/renderer/src/features/terminal/TerminalView.tsx`**
- New `__clearTerminalSelection` test hook: clears xterm selection plus the `confirmedSelectionRef` / `confirmedDomSelectionRef` / `extendedSelectionRef` fallbacks (plain `clearSelection()` left stale values behind via the `readTerminalSelection()` fallback chain). Removed on unmount with the other hooks.
- Justification comment for unconditional test hooks: packaged E2E runs against the production build with no test-flag plumbing, and the same-origin renderer already owns the terminal instance.

## Invariants

- **Source of truth**: xterm buffer (`buffer.active`) for output/marker/selection state; stub gateway for inputs/resizes.
- **Lock/transaction scope**: none — single-page E2E with sequential cases and a `beforeEach` reset.
- **Acceptable orphan states**: a failed gesture can leave prior selection/clipboard behind; guarded by sentinels, pre-copy selection asserts, and diagnostics/screenshots.
- **Auth source of truth**: stub `auth:start-device-flow` device flow (test-only).
- **Deferred scope**: full per-case page isolation; repeated Xvfb stability runs (needs CI).

## Surface matrix

N/A on all surfaces — test-only change; the `TerminalView` delta is test hooks with no visual or behavioral effect.

| Surface | UI | Behavior | State/recovery | Automated tests | Real evidence |
| --- | --- | --- | --- | --- | --- |
| Web Canvas | N/A | N/A | N/A | N/A | N/A |
| Web Desktop | N/A | N/A | N/A | N/A | N/A |
| Electron Desktop | N/A | N/A | N/A | N/A | N/A |
| Web Mobile | N/A | N/A | N/A | N/A | N/A |
| Native Mobile | N/A | N/A | N/A | N/A | N/A |

## Tests

- [x] `npx react-doctor@latest desktop` — ran; the 5 `TerminalView.tsx` hits are pre-existing lines (theme-ref write, session-change ref clears), zero new findings from this change
- [x] Manual pattern sweep (`grep`) on both changed files — no bare catches, no untimed `fetch`, no sync file I/O, no unbounded `Map`/`Set`
- [x] `bun build` syntax check on both files — only missing-dependency resolutions (no `node_modules` in this env), no syntax errors

### DIFF
--- a/tests/e2e/desktop/terminal-clipboard.e2e.test.ts
+++ b/tests/e2e/desktop/terminal-clipboard.e2e.test.ts
@@ -874,6 +874,8 @@ suite("packaged Electron production-mode terminal selection", () => {
         screenBox.y + screenBox.height / 2,
       );
       await page.mouse.wheel(0, -100_000);
+      await expect.poll(async () => (await readTerminalDiagnostics()).viewportY).toBe(0);
+      await waitForRenderFrames(2);
       const clickTop = await expectTerminalPoint(point(2, 2), "mouse-edge-top-click");
       await page.mouse.click(clickTop.x, clickTop.y);
       gateway.sendTerminalOutput("\u001b[?1003h\u001b[?1006h");
--- a/tests/e2e/desktop/terminal-clipboard.e2e.test.ts
+++ b/tests/e2e/desktop/terminal-clipboard.e2e.test.ts
@@ -2,7 +2,7 @@ import { existsSync, mkdirSync, mkdtempSync, rmSync } from "node:fs";
 import { createRequire } from "node:module";
 import { tmpdir } from "node:os";
 import { join, resolve } from "node:path";
-import { afterAll, beforeAll, describe, expect, it } from "vitest";
+import { afterAll, beforeAll, beforeEach, describe, expect, it } from "vitest";
 import { _electron, type ElectronApplication, type Locator, type Page } from "playwright";
 import { startStubGateway, type StubGateway } from "./fixtures/stub-gateway";
 
@@ -232,6 +232,10 @@ suite("packaged Electron terminal clipboard", () => {
     await page.keyboard.press(copyShortcut);
     await expect.poll(clipboardText).toContain(firstLine);
     const selectAllSnapshot = await clipboardText();
+    // Intentional stability soak (not output-readiness sync): with mouse
+    // reporting enabled, roam the pointer to prove stray motion reports do
+    // not disturb the Select All snapshot. Keep the fixed dwell so each of
+    // the 20 positions gets time to emit before the re-copy assertion.
     for (let move = 0; move < 20; move += 1) {
       await page.mouse.move(
         screenBox.x + 10 + ((move * 31) % Math.max(20, screenBox.width - 20)),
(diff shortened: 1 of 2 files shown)


tests/e2e/desktop/terminal-clipboard.e2e.test.ts
```diff
@@ -533,59 +833,85 @@ suite("packaged Electron production-mode terminal selection", () => {
...
 
       gateway.sendTerminalOutput("\u001b[?1003l\u001b[?1006l");
+      await waitForMouseTrackingMode("none");
       await page.mouse.move(
         screenBox.x + screenBox.width / 2,
         screenBox.y + screenBox.height / 2,
       );
       await page.mouse.wheel(0, -100_000);
-      await page.mouse.click(point(2, 2).x, point(2, 2).y);
+      const clickTop = await expectTerminalPoint(point(2, 2), "mouse-edge-top-click");
```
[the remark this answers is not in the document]
sundaram2021: suggesstion applied

Nima-Naderi:
@greptileai can you review again

tests/e2e/desktop/terminal-clipboard.e2e.test.ts
```diff
@@ -306,6 +310,206 @@ suite("packaged Electron production-mode terminal selection", () => {
...
+
+  async function waitForMouseTrackingMode(
+    expected: "none" | "!none",
+    options?: { timeout?: number },
+  ): Promise<void> {
+    await expect.poll(async () => {
+      const diag = await readTerminalDiagnostics();
+      return expected === "none"
+        ? diag.mouseTrackingMode === "none"
+        : diag.mouseTrackingMode !== "none";
```
HamedMP: **P3 — Require a known enabled mouse mode.** `readTerminalDiagnostics()` defaults this value to `"unknown"`, so the `!== "none"` branch can pass before mouse tracking is actually observed. Please accept only the concrete non-`none` modes emitted by the terminal diagnostics.

tests/e2e/desktop/terminal-clipboard.e2e.test.ts
```diff
@@ -306,6 +310,206 @@ suite("packaged Electron production-mode terminal selection", () => {
...
+      );
+    }
+    return point;
+  }
+
+  async function dismissOpenContextMenu(): Promise<void> {
+    const menu = page.getByRole("menu", { name: "Terminal actions" });
+    if (await menu.isVisible().catch(() => false)) {
+      await page.keyboard.press("Escape");
+      await menu.waitFor({ state: "hidden", timeout: 5_000 }).catch(() => null);
```
HamedMP: **P2 — Do not continue after menu dismissal fails.** Swallowing this timeout allows the next test step to run with the Terminal actions menu still open, which can intercept pointer input and recreate the flake. Let the wait fail, or log the diagnostic and rethrow.

tests/e2e/desktop/terminal-clipboard.e2e.test.ts
```diff
@@ -347,24 +551,31 @@ suite("packaged Electron production-mode terminal selection", () => {
...
       ), { x: screenBox.x + 2, y })).toBe(true);
     }
     return {
       screenBox,
-      resize,
+      resize: { cols, rows },
       point: (column: number, row: number) => ({
-        x: screenBox.x + (column + 0.25) * (screenBox.width / resize.cols),
-        y: screenBox.y + (row + 0.5) * (screenBox.height / resize.rows),
+        x: screenBox.x + (column + 0.5) * (screenBox.width / cols),
```
HamedMP: **P2 — Avoid the xterm fractional rounding boundary.** Centering at `column + 0.5` lands on the exact boundary affected by xterm's fractional-cell rounding. Restore the safer fractional offset or use explicit start/end offsets so selection endpoints do not round into adjacent cells.

tests/e2e/desktop/terminal-clipboard.e2e.test.ts
```diff
@@ -401,84 +612,103 @@ suite("packaged Electron production-mode terminal selection", () => {
...
-      await page.mouse.click(point(2, 0).x, point(2, 0).y);
+      const probePoint = await expectTerminalPoint(point(2, 0), "mouse-mode-probe-click");
+      await page.mouse.click(probePoint.x, probePoint.y);
       await expect.poll(() => gateway.state.terminalInputs.length).toBeGreaterThan(beforeProbeClick);
       expect(gateway.state.terminalInputs.slice(beforeProbeClick).some((data) => data.includes("\u001b[<")))
         .toBe(true);
 
       await page.mouse.move(start.x, start.y);
-      await page.waitForTimeout(100);
+      await waitForRenderFrames(1);
```
HamedMP, line 738: **P2 — Synchronize at the gateway boundary before recording the drag baseline.** A renderer animation frame does not prove the mouse-motion report from `page.mouse.move(start)` has reached the stub gateway. That legitimate report can arrive after `beforeDrag` and make the negative assertion fail. Wait for the expected gateway report or quiescence before taking the baseline, and use an observable transport boundary after the drag as well.

HamedMP (commented):
Manual review found four remaining E2E-stability issues that should be addressed before merge.

sundaram2021:
fixed all the  review issues
- P3 mouse modes:  — !none now accepts only x10|vt200|drag|any; beforeEach uses same allowlist; "unknown" correctly keeps waiting.
- P2 menu dismissal: — swallowing .catch replaced with diagnostics + screenshot + rethrow with cause.
- P2 rounding boundary: — x-offset back to column + 0.25, consistent with existing selectBetween anchors.
- P2 gateway sync: — waitForGatewayInputsToSettle() (300ms quiet window, 5s timeout) called before beforeDrag baseline and after drag before the empty assertion.

Issue #1676, opened 2026-09-15
chore(arch): codebase architecture audit — de-monolithization, AI unification, and typed SDK


sundaram2021:
## Summary

Matrix OS ("Web 4": headless AI OS + multi-tenant control plane + per-owner VPS runtime) has a clear vision in `ARCHITECTURE.md` (headless core, owner-owned data, multi-shell renderers), but rapid feature growth has produced severe structural debt:

- 36+ files > 1,000 LOC, with entry points up to ~5,300 LOC, violating our own `<500 LOC` budget in `docs/dev/large-file-refactoring.md`.
- 123 flat files in `packages/gateway/src/` root + 127 flat files in `packages/platform/src/` root (Spec 093 unfulfilled — grew from 86 since June 2026).
- 4 parallel AI execution tracks (`kernel` + `gateway/chat/` + `gateway/coding-agents/` + `symphony-elixir/`) with no shared session/turn model.
- 3 independent conversation stores (JSON files vs. Postgres `chats` vs. coding-agent thread store).
- 4 hand-written Gateway clients (`shell`, `apps/mobile`, `packages/sync-client`, `desktop`) duplicating the same endpoints (~4,300+ LOC boilerplate).
- Boundary violations: Gateway → Platform DB direct connection; ATS hiring tool inside Platform control plane.
- 8-layer Terminal/Zellij stack held together by 6 separate recovery daemons.
- Product split-brain: Canvas vs. Classic Desktop, Chat vs. Coding Agent vs. Terminal silos, 7-step VPS onboarding, iframe miniapps vs. real project previews.

This aligns with `docs/dev/pr-review-analysis.md` (~317 unresolved review comments, ~1.5% resolution rate): violations of the mandatory patterns in `AGENTS.md` are the #1 bug source.

This issue tracks the audit findings, the proposed solution workstreams, and acceptance criteria. 

## Current Architecture (for context)

```mermaid
flowchart TB
    subgraph Entry["Entry: matrix-os.com / app.matrix-os.com"]
        Site["matrix-os-site<br/>(separate repo)<br/>marketing + signup"]
        EdgeRouter["packages/edge-router/<br/>Cloudflare routing helpers"]
        NeoWorker["packages/neo-worker/<br/>PostHog / edge support"]
    end

    subgraph Clients["Client Layer (renderers, no canonical truth)"]
        WebShell["shell/<br/>Next.js 16 web shell<br/>Canvas + Desktop + stores"]
[...]

PR #1680, opened 2026-09-15
chore(arch): add architecture budget gate for #1676 Phase 0

sundaram2021:
Part of the phased fix plan for HamedMP/matrix-os#1676 — this PR implements **Phase 0 (guardrails + baseline)**. No behavior change; it adds a CI-enforceable ratchet so structural debt can only shrink from here.

## Summary

- Adds `bun run check:budgets` gate: pinned file-LOC ratchets, 1000-LOC cap with allowlist, and flat-directory caps for `packages/gateway/src` + `packages/platform/src`
- Baselines verified against `main @ 8a250121`: 38 ratcheted files, dir caps 123 / 127, every allowlist entry tagged with its owning phase
-  Hooks the gate into `check:patterns.sh` as Pass 9 (always full-tree, documented why)

## Fixes applied

-  New god files (>1000 LOC, not allowlisted) now fail CI instead of silently landing
-  Pinned files (`server.ts`, `platform/db.ts`, …) cannot grow past recorded LOC — values only move down
-  Flat roots cannot gain files (Spec 093 sprawl guard: gateway 123, platform 127)
- Review feedback: reject non-directory/missing scan roots; documented full-tree Pass 9 rationale

## Files affected

- `scripts/review/check-budgets.mjs` (new) — gate implementation + CLI
- `scripts/review/arch-budgets.json` (new) — verified baseline + phase-tagged allowlist
- `tests/scripts/check-budgets.test.ts` (new) — 9 vitest cases
- `scripts/review/check-patterns.sh` — Pass 9 hook (+ full-tree rationale comment)
- `package.json` — `check:budgets` script

## Tests ran

- [x] `vitest run tests/scripts/check-budgets.test.ts` — 9/9 pass (TDD: Red → Green)
- [x] `node scripts/review/check-budgets.mjs` on real tree — 0 violations, 38 tracked warnings
- [x] CLI contract — exit 0 clean / 1 violation / 2 bad-args (verified with fixtures)
- [x] `bash -n scripts/review/check-patterns.sh` — syntax clean

## Invariants

- **Source of truth**: `scripts/review/arch-budgets.json`; ratchet values only move down, entries pruned in the same PR that deletes/splits the file
- **Lock/transaction scope**: N/A (read-only scan, no writes)
- **Acceptable orphan states**: N/A (stateless CLI; stale entries degrade to warnings, never failures)
- **Auth source of truth**: N/A (local review tooling, no endpoints)
- **Deferred scope**: full `check:patterns` (needs `rg`) and repo-wide `typecheck` were not runnable in this env — no TS files touched; CI is the backstop. Product phases 1–6 per #1676 plan land as follow-up PRs.


sundaram2021:
fixed review issues

scripts/review/arch-budgets.json
```diff
@@ -0,0 +1,88 @@
+{
+  "$comment": "Architecture budgets for issue #1676 (Phase 0 guardrails). Verified against origin/main @ 8a250121. Numbers only move DOWN; prune entries in the same PR that deletes/splits the file. Allowlist entries require a linked phase owner.",
+  "fileCap": 1000,
```
HamedMP, line 3: **P1 — Baseline the gate against the current PR tree.** Running this checker at head `904d644b` produces 17 violations: for example `db.ts` is 5316 > 5301, `server.ts` is 4859 > 4769, and the gateway/platform flat-root counts are 129 > 123 and 130 > 127. There are also existing unallowlisted files above the cap, including `collaboration/routes.ts` and the generated agent-inspirations file. Reconcile the baseline and generated-file treatment against current `main`, then add a test that runs the committed config against the real repository tree.

scripts/review/check-patterns.sh
```diff
@@ -233,6 +233,33 @@ if [[ -n "$MATCHES" ]]; then
...
 
+# ═══════════════════════════════════════════════════════════════
+# PASS 3: Architecture budgets (issue #1676 Phase 0)
+# ═══════════════════════════════════════════════════════════════
+
+# NOTE: budgets always scan the whole tree, even in --diff mode. Ratchets
+# and dir caps are absolute (not per-diff), so a diff-scoped run would miss
+# growth elsewhere. Cost is a few seconds; do not "optimize" this to --diff.
+header "9. Architecture budgets — file LOC ratchets and flat-dir caps"
+
```
HamedMP: **P2 — Run the architecture gate before the diff scanner's early exit.** CI invokes this script with `--diff`, but lines 55–59 exit successfully when no TypeScript file changed under `packages/` or `shell/`. An oversized file added only under `desktop/`, `apps/mobile/`, or `home/apps/` therefore skips this supposedly full-tree check. Move the budget invocation before that return or run it as an independent CI step, and cover this case with an integration test.

HamedMP (commented):
The gate currently starts red against this PR's own tree and can also be bypassed by diff-only CI. Please reconcile both issues and add integration coverage before enabling it.

PR #1762, opened 2026-09-19
fix(codex): verify provider contracts for 0.155.1

xyhStruggler:
## Summary

`Codex Provider Contracts` fails because the published latest version is 0.155.1 while Matrix only verifies through 0.154.0. Record the reviewed exec/app-server schema hashes and align managed Codex installations with 0.155.1. The workflow continues to check the actual latest release.

The exec schema is unchanged. App-server adds memory/status, thread attachment, and verification-cancellation methods plus an optional feedback field; every method and notification consumed by Matrix retains its semantic digest. Existing verified versions remain supported.

## Validation

- Generated the official 0.154.0 and 0.155.1 macOS app-server schemas and compared the contract surfaces; the Linux CI failure reported the same 0.155.1 schema digest.
- Provider contracts, event adapters, contract checker, and elicitation regressions: 41 passed.
- Host installer/package checks: 71 passed.
- Typecheck passed; pattern scan reported zero violations.
- Full repository suite on macOS: 16,026 passed, 29 failed, 50 skipped. Every failing case is independently reproduced on untouched main in focused comparison runs, including the three interactive installer timeouts; no clean local full-suite claim.
- Both Linux x64 and macOS arm64 official provider-contract CI jobs pass. Greptile reviewed this head at 5/5. Full repository CI remains required before merge.

## Invariants

- Source of truth: versioned provider contracts and `CODEX_VERIFIED_VERSION`; published schemas must match recorded hashes and consumed semantic digests.
- Lock/transaction scope: unchanged; no database or runtime mutation paths change.
- Acceptable orphan states: none introduced.
- Auth source of truth: existing provider authentication and approval policy remain unchanged; no new upstream capability receives a grant.
- Deferred scope: no support is added for the new optional upstream methods.


xyhStruggler:
[codex] @greptileai review — verified published Codex 0.155.1 schemas, retained the strict latest-version guard, and aligned managed installation pins. Consumed method/notification semantic digests are unchanged.

PR #1801, opened 2026-09-21
fix(ci): restore main runtime validation

HamedMP:
Main CI currently fails for two deterministic repository-owned reasons: the Docker gateway starts before `@matrix-os/terminal-runtime` has emitted its `dist` exports, and the platform schema baseline omits two funded-usage indexes already present in the migration.

This builds and assigns ownership of the terminal runtime output before gateway startup, adds a regression contract for the build order, and updates the schema fixture to the live index set.

Validation:
- `tests/scripts/docker-dev-entrypoint.test.ts`: 4/4 passed
- `tests/platform/collaboration-foundation.test.ts`: 8/8 passed
- Clean-checkout gateway import reproduced `ERR_MODULE_NOT_FOUND` before the terminal runtime build and succeeded afterward
- `sh -n distro/docker-dev-entrypoint.sh`
- `git diff --check`


HamedMP:
@greptile review

HamedMP:
@greptile review

